PR #10, opened 2020-03-30
Release jars via github actions

phact:
Although users are most likely to use the management API via docker or k8s, if someone wants to use it with Cassandra standalone we should provide the jars in our github releases.

This PR includes the actions workflow that creates releases based on tags and uploads the jars to the release as `jars.zip`.


phact:
Btw, I tested this in my fork https://github.com/phact/management-api-for-apache-cassandra/releases/tag/v1.0.3

PR #15, opened 2020-05-18
CDB-34 Added Docker files and scripts to make the management API work…

sbtourist:
… with Astra-flavoured OSS distributions.

### DIFF
--- a/astra-4.0/Dockerfile
+++ b/astra-4.0/Dockerfile
@@ -1,0 +1,105 @@
+FROM adoptopenjdk:8-jre-hotspot-bionic
+
+# explicitly set user/group IDs
+RUN set -eux; \
+	groupadd -r cassandra --gid=999; \
+	useradd -r -g cassandra --uid=999 cassandra
+
+RUN set -eux; \
+	apt-get update; \
+	apt-get install -y --no-install-recommends \
+# solves warning: "jemalloc shared library could not be preloaded to speed up memory allocations"
+		libjemalloc1 \
+# "free" is used by cassandra-env.sh
+		procps \
+# "cqlsh" needs a python interpreter
+		python \
+# "ip" is not required by Cassandra itself, but is commonly used in scripting Cassandra's configuration (since it is so fixated on explicit IP addresses)
+		iproute2 \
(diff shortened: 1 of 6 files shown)


astra-4.0/Dockerfile
```diff
@@ -0,0 +1,105 @@
+FROM adoptopenjdk:8-jre-hotspot-bionic
```
sbtourist, line 1: @stef1927 FYI I've amended this back to a jre image, same as the upstream OSS one (while I previously switched to a jdk image for a test).

sbtourist:
Thanks @stef1927. @tjake and/or @jimdickinson, any interest in reviewing this?

tjake:
Do you want to publish this to docker as well if tagged?

tjake:
https://github.com/datastax/management-api-for-apache-cassandra/blob/master/.github/workflows/docker-release.yaml

sbtourist:
> Do you want to publish this to docker as well if tagged?

@tjake It is currently tagged as 4.0, but such tag would be continuously updated as we develop it. I'm not sure that's a good practice. I'm also wondering if it makes sense to release it as "4.0" given we're soon going to diverge significantly from OSS 4.0. I'd say not to publish it until we figure things out.

scripts/docker-entrypoint.sh
```diff
@@ -70,36 +70,66 @@ if [ "$1" = 'mgmtapi' ]; then
...
-	# for rackdc in dc rack; do
-	# 	var="CASSANDRA_${rackdc^^}"
-	# 	val="${!var}"
-	# 	if [ "$val" ]; then
-	# 		_sed-in-place "$CASSANDRA_CONF/cassandra-rackdc.properties" \
-	# 			-r 's/^('"$rackdc"'=).*/\1 '"$val"'/'
-	# 	fi
-	# done
+    # Set this if you want to ignore default env variables, i.e. when running inside an operator
+    if [ $IGNORE_DEFAULTS ]; then
```
jimdickinson, line 74: this needs to be the default way this image works for compatibility with published operator versions

PR #24, opened 2020-06-24
CDB-40: Use JDK11 instead of 1.8 for the mgmt API

nastra:


management-api-agent/pom.xml
```diff
@@ -170,8 +170,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.5.1</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>11</source>
+                    <target>11</target>
```
johntrimble, line 174: Won't this cause a problem because we support Cassandra 3.11 and it only supports Java 8, doesn't it?

nastra: yes, this causes problems and so we can't really use JDK11 for the project itself unfortunately

PR #27, opened 2020-07-02
Add API endpoint that allows creating a Keyspace

nastra:


management-api-agent/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -318,4 +321,18 @@ public void createRole(@RpcParam(name="username") String username,
     {
         return ShimLoader.instance.get().getStreamInfo();
     }
+
+    @Rpc(name = "createKeyspace")
+    public void createKeyspace(String keyspaceName, Map<String, String> replicationSettings) throws IOException
+    {
+        logger.debug("Creating keyspace {} with replication settings {}", keyspaceName, replicationSettings);
+        Map<String, Integer> settings = new HashMap<>();
```
RazakDangi, line 329: [minor] using map(transforming) would stop us from creating new local map with new
		
Map<String, Integer> settings = replicationSettings.entrySet().stream()
				.collect(
						Collectors.toMap(Map.Entry::getKey, entry ->Integer.valueOf(entry.getValue()))
				);

management-api-agent/pom.xml
```diff
@@ -69,6 +69,11 @@
                     <version>4.11</version>
                     <scope>test</scope>
                 </dependency>
+                <dependency>
```
RazakDangi, line 72: [minor] I would keep compile dependencies above test scope dependencies

nastra: fixed

management-api-agent/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
nastra: I was only checking something with this code, but yes, the replication settings will usually always be a pair of String, Integer

nastra:
> I think the operator also needs an alter keyspace replication for when it brings up a new DC.

@tjake np I can add that. Maybe as a separate issue/PR?

PR #30, opened 2020-07-06
Add API endpoint that allows altering the RF settings of a Keyspace

nastra:


management-api-server/src/main/java/com/datastax/mgmtapi/resources/KeyspaceOpsResources.java
```diff
@@ -94,50 +95,58 @@ public Response refresh(@QueryParam(value="keyspaceName")String keyspaceName, @Q
...
+        });
     }
 
     @POST
     @Path("/create")
     @Produces(MediaType.TEXT_PLAIN)
     @Consumes(MediaType.APPLICATION_JSON)
     @Operation(summary = "Create a new keyspace with the given name and replication settings")
-    public Response create(CreateKeyspaceRequest createKeyspaceRequest)
+    public Response create(CreateOrAlterKeyspaceRequest createOrAlterKeyspaceRequest)
```
RazakDangi, line 106: We could make actions as enum type 
KeySpaceActionEnum{
CREATE("create"),
REFRESH("refresh),
CLEANUP("cleanup"),
ALTER("alter");}
As we dont have to see them as different resource but as single endpoint with different Actions
We save duplicate code , avoid multiple endpoints

nastra: Thanks for your review @RazakDangi. It's not really in the scope of this PR to refactor all of those things. Would you mind opening up an issue that describes your suggestion in more detail? Also feel free to contribute a patch with your suggestion.

PR #36, opened 2020-07-28
Experimental arm64 support for Cassandra 3.11.6 

### LINKED ISSUE #35: Add experimental arm64 support for Cassandra 3.11.6
The official Cassandra [images](https://hub.docker.com/_/cassandra?tab=tags) all support a variety of platforms including arm64. Support for arm64 is important for users who are using Cassandra in an embedded/IoT setting, are developing on their new ARM powered Apple laptop, or are looking to save money by using arm servers, such as with AWS Graviton. Ideally, we would align with OSS Cassandra to better enable these users.

Supporting arm64 comes with a number of challenges:

1. Metrics collector does not presently support arm
2. There is no official Cassandra 4.0 image and ours does not support arm
3. Netty only provides arm64 builds for epoll support starting with 4.1.50
4. We have no arm hardware for building
5. We do not presently have arm hardware for testing

This ticket is narrowly focused on just getting Cassandra 3.11.6 to work, in any capacity, on arm64 with the Management API. 3.11.6 already has an official multiarch image and the Cass Operator does testing against 3.11.6 regularly, so it's a good candidate to start with.

For this ticket, I propose we address the above issues in the following way:

1. Disable the metrics collector when running on arm
2. Ignore Cassandra 4.0 for the moment
3. Update the version of Netty for the management API
4. We can use [buildx](https://docs.docker.com/buildx/working-with-buildx/), which will in turn use QEMU, to produce builds for arm without the need of arm hardware
5. I can test on a raspberry pi, but it's not ideal. I think its worth it to get an arm build out there even if we cannot test it regularly.

johntrimble:
Turns the Cassandra 3.11.6 w/ Management API built into a multiarch build supporting both arm64 and amd64. Currently, disables metrics collector on arm64 as our collectd fork does not presently support arm64.

Closes #35

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,11 +18,6 @@ jobs:
           path: ~/.m2
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
-      - name: Setup Buildx
-        id: buildx
-        uses: crazy-max/ghaction-docker-buildx@v3
-        with:
-          version: latest
       - name: Build Management API in docker
         run: |
           cat <<EOF > ~/.m2/settings.xml
@@ -44,11 +39,7 @@ jobs:
           cp ~/.m2/settings.xml settings.xml
           docker build -t management-api-for-dse-builder -f ./Dockerfile-build-dse ./
           docker tag management-api-for-dse-builder management-api-for-apache-cassandra-builder
-          docker buildx build --load \
-            --tag mgmtapi-3_11 \
-            --file Dockerfile-oss \
-            --target oss311 \
-            --platform linux/amd64 .
+          docker build -t mgmtapi-3_11 -f Dockerfile-3_11 .
           docker build -t mgmtapi-4_0 -f Dockerfile-4_0 .
           docker build -t mgmtapi-dse-68 -f Dockerfile-dse-68 .
       - name: Build with Maven
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,6 +18,11 @@ jobs:
           path: ~/.m2
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
+      - name: Setup Buildx
+        id: buildx
+        uses: crazy-max/ghaction-docker-buildx@v3
+        with:
+          version: latest
       - name: Build Management API in docker
         run: |
           cat <<EOF > ~/.m2/settings.xml
@@ -39,7 +44,11 @@ jobs:
           cp ~/.m2/settings.xml settings.xml
           docker build -t management-api-for-dse-builder -f ./Dockerfile-build-dse ./
           docker tag management-api-for-dse-builder management-api-for-apache-cassandra-builder
-          docker build -t mgmtapi-3_11 -f Dockerfile-3_11 .
+          docker buildx build --load \
+            --tag mgmtapi-3_11 \
+            --file Dockerfile-oss \
+            --target oss311 \
+            --platform linux/amd64 .
           docker build -t mgmtapi-4_0 -f Dockerfile-4_0 .
           docker build -t mgmtapi-dse-68 -f Dockerfile-dse-68 .
       - name: Build with Maven
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,11 +44,11 @@ jobs:
           cp ~/.m2/settings.xml settings.xml
           docker build -t management-api-for-dse-builder -f ./Dockerfile-build-dse ./
           docker tag management-api-for-dse-builder management-api-for-apache-cassandra-builder
-          docker buildx build --load \
+          docker buildx build \
             --tag mgmtapi-3_11 \
             --file Dockerfile-oss \
             --target oss311 \
-            --platform linux/amd64 .
+            --platform linux/amd64,linux/arm64 .
           docker build -t mgmtapi-4_0 -f Dockerfile-4_0 .
           docker build -t mgmtapi-dse-68 -f Dockerfile-dse-68 .
       - name: Build with Maven
--- a/Dockerfile-oss
+++ b/Dockerfile-oss
@@ -20,15 +20,18 @@ COPY management-api-shim-4.x ./management-api-shim-4.x
 RUN mvn -q -ff package -DskipTests
 
 FROM --platform=$BUILDPLATFORM maven:3.6.3-jdk-8-slim as netty4150
-RUN mvn dependency:get -DgroupId=io.netty -DartifactId=netty-all -Dversion=4.1.50.Final -Dtransitive=false
+RUN mvn dependency:get -DgroupId=io.netty -DartifactId=netty-transport-native-epoll -Dversion=4.1.50.Final -Dclassifier=linux-aarch64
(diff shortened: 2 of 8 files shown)


Dockerfile-oss
```diff
@@ -0,0 +1,115 @@
+FROM --platform=$BUILDPLATFORM maven:3.6.3-jdk-8-slim as builder
```
johntrimble, line 1: When using buildx, there isn't a way to make a build of an image in one invocation of buildx available to a subsequent invocation of buildx without the use of a registry. This means our traditional approach of having a 'build' docker image that we then reference in subsequent docker files will not work (unless we push it to a registry). To get around this, we use one multi-stage docker file that contains the build stages. A further discussion of this issue in buildx can be found here: https://github.com/docker/buildx/issues/156

Dockerfile-oss
```diff
@@ -0,0 +1,115 @@
...
+FROM --platform=$BUILDPLATFORM maven:3.6.3-jdk-8-slim as netty4150
+RUN mvn dependency:get -DgroupId=io.netty -DartifactId=netty-all -Dversion=4.1.50.Final -Dtransitive=false
+
+FROM --platform=linux/amd64 cassandra:3.11 as oss311-amd64
+
+FROM --platform=linux/arm64 cassandra:3.11 as oss311-arm64
+# Netty arm64 epoll support was not added until 4.1.50 (https://github.com/netty/netty/pull/9804)
+# Only replace this dependency for arm64 to avoid regressions
+RUN rm /opt/cassandra/lib/netty-all-*.jar
+COPY --from=netty4150 /root/.m2/repository/io/netty/netty-all/4.1.50.Final/netty-all-4.1.50.Final.jar /opt/cassandra/lib/netty-all-4.1.50.Final.jar
```
johntrimble, line 31: The better way to do this is probably to have the management api agent contain a shaded version of netty at the appropriate version. That will take further experimentation. However, we are only swapping out this jar for arm64, so this should not impact any existing users.

Dockerfile-oss
johntrimble: Also, note the `--platform=$BUILDPLATFORM`. This is because we don't have to build the jar for each arch we support since its java. This indicates that this stage should only be built on whatever the build platform is (probably linux-amd64).

Dockerfile-oss
```diff
@@ -0,0 +1,115 @@
...
+ARG TARGETARCH
+
+COPY --from=builder /build/management-api-common/target/datastax-mgmtapi-common-0.1.0-SNAPSHOT.jar /etc/cassandra/
+COPY --from=builder /build/management-api-agent/target/datastax-mgmtapi-agent-0.1.0-SNAPSHOT.jar /etc/cassandra/
+COPY --from=builder /build/management-api-server/target/datastax-mgmtapi-server-0.1.0-SNAPSHOT.jar /opt/mgmtapi/
+COPY --from=builder /build/management-api-shim-3.x/target/datastax-mgmtapi-shim-3.x-0.1.0-SNAPSHOT.jar /opt/mgmtapi/
+COPY --from=builder /build/management-api-shim-4.x/target/datastax-mgmtapi-shim-4.x-0.1.0-SNAPSHOT.jar /opt/mgmtapi/
+
+ENV TINI_VERSION v0.18.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-${TARGETARCH} /tini
```
johntrimble, line 44: Note the change here to get tini for the appropriate architecture.

Dockerfile-oss
```diff
@@ -0,0 +1,115 @@
...
+EXPOSE 8080
+
+ENTRYPOINT ["/docker-entrypoint.sh"]
+CMD ["mgmtapi"]
+
+
+# NOTE: Presently, OSS doesn't have official Cassandra 4.0 builds on dockerhub
+# and our build at datastax/cassandra:4.0 is not a multiarch image like the
+# official ones. Once one of those issues is fixed, the following targets can
+# be used to build Cassandra 4.0 with Management API.
```
johntrimble, line 73: Half way through this I realized we didn't have an appropriate multiarch cassandra 4.0 image. We could just remove all of this for now instead of commenting out. I thought it might be useful for reference, but I have no strong feelings either way.

management-api-common/pom.xml
```diff
@@ -14,7 +14,7 @@
     <properties>
         <slf4j.version>1.7.25</slf4j.version>
         <logback.version>1.2.3</logback.version>
-        <netty.version>4.1.45.Final</netty.version>
+        <netty.version>4.1.50.Final</netty.version>
```
johntrimble, line 17: This one provides a build of epoll for arm64.

scripts/docker-entrypoint.sh
```diff
@@ -40,6 +40,11 @@ _sed-in-place() {
 	rm "$tempFile"
 }
 
+_metrics_collector_supported() {
+	# currently, metrics collector does not work on arm64
+	[ "$(uname -m)" != "aarch64" ]
+}
```
johntrimble, line 46: I don't know the best way to handle this. Somehow we need to disable the metrics collector for arm64. This appears to be working, but I'm open to suggestions.

johntrimble:
@tjake yeah, I created the branch in the datastax repo, and since both branches pointed to the same commit, the jobs just started running. Tests are passing now, but I'm putting this back into draft briefly to test something.

.github/workflows/ci.yaml
```diff
@@ -39,7 +44,11 @@ jobs:
           cp ~/.m2/settings.xml settings.xml
           docker build -t management-api-for-dse-builder -f ./Dockerfile-build-dse ./
           docker tag management-api-for-dse-builder management-api-for-apache-cassandra-builder
-          docker build -t mgmtapi-3_11 -f Dockerfile-3_11 .
+          docker buildx build --load \
+            --tag mgmtapi-3_11 \
+            --file Dockerfile-oss \
+            --target oss311 \
+            --platform linux/amd64 .
```
johntrimble, line 51: Only building here for "linux/amd64" as a multiarch image cannot be loaded into the local docker images. This should be good enough for running the integration tests.

johntrimble:
@tjake The ci tests are passing now, but that's a bit deceptive. I just noticed that we use a docker-java library to build the image in our integration tests:

https://github.com/datastax/management-api-for-apache-cassandra/blob/cassandra-mgmtapi-multiarch-pr/management-api-server/src/test/java/com/datastax/mgmtapi/helpers/DockerHelper.java#L203-L208

Currently, this is using the Dockerfile-3_11 docker file. However, I updated the build to use Dockerfile-oss to create the multiarch images:

https://github.com/datastax/management-api-for-apache-cassandra/blob/cassandra-mgmtapi-multiarch-pr/.github/workflows/docker-release.yaml#L62-L69

This dockerfile requires the use of buildkit, which one can use by either using buildx (as is done to generate the multiarch images) or by setting the environment variable `DOCKER_BUILDKIT=1` when calling `docker build`. Unfortunately, this docker-java library has no support for buildx and doesn't seem to have any intention of supporting buildkit: https://github.com/docker-java/docker-java/issues/1381

Is there any reason why we cannot just invoke the docker executable to do the build instead of using this library?

tjake:
> Is there any reason why we cannot just invoke the docker executable to do the build instead of using this library?

That should be fine, can you do that?

johntrimble:
> That should be fine, can you do that?

Yes, I'll take care of it. Just wanted to make sure it was palatable first. Thanks for the feedback!

johntrimble:
@tjake Okay, integration tests are now passing! My Java is a little rusty, so let me know if I should be doing anything differently there.

PR #56, opened 2020-12-04
Update jar-release action

emerkle826:
This PR actually addresses two small issues:

- The jar-release action uses `setup-java`, but uses an old version (v1.3.0) that uses `set-env` and `set-path`, which are not allowed any more. This PR updates to <v1> which is what is used in other actions and is actually the latest
- In merging #51, I missed the fact that `Dockerfile-astra-4_0` and `Dockerfile-oss` also needed similar changes

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,7 +48,12 @@ jobs:
             --tag mgmtapi-3_11 \
             --file Dockerfile-oss \
             --target oss311 \
-            --platform linux/amd64,linux/arm64 .
+            --platform linux/amd64 .
+          docker buildx build \
+            --tag mgmtapi-3_11 \
+            --file Dockerfile-oss \
+            --target oss311 \
+            --platform linux/arm64 .
           docker build -t mgmtapi-4_0 -f Dockerfile-4_0 .
           docker build -t mgmtapi-dse-68 -f Dockerfile-dse-68 .
       - name: Build with Maven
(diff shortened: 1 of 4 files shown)


emerkle826:
The CI build has been failing on this branch, but I'm not sure what the issue is. It fails with:
```
failed to solve: rpc error: code = Canceled desc = failed to compute cache key: failed to copy: httpReadSeeker: failed open: failed to do request: context canceled
```
Inspection of the build logs show that a few of the Dockerfile build steps are "CANCELED", which I suspect is related. However I've tried the exact same build steps locally and it passes. I've tried removing `linux/arm64` from the ci.yaml (essentially disabling multi-platform building) and the build passes this point.

.github/workflows/ci.yaml
```diff
@@ -48,7 +48,12 @@ jobs:
             --tag mgmtapi-3_11 \
             --file Dockerfile-oss \
             --target oss311 \
-            --platform linux/amd64,linux/arm64 .
+            --platform linux/amd64 .
```
emerkle826, line 51: @johntrimble and/or @tjake I would appreciate your thoughts on this. It seems sometime since I tagged v0.1.14, the multi-platform build step fails with a "context canceled" error, and I haven't been able to figure out why. Splitting up the build into 2 separate builds seems to work. I have tried this locally on a Linux machine with the docker binfmt extras for building ARM targets, and it works as expected without splitting it up, but it seems to fail on GitHub.

johntrimble: It's puzzling to me that this would fail, but if releases are still working, I'm not too concerned about having to split them up here. Splitting these like this when doing the release would cause problems (we'd end up with a single arch build and the last one built would win).

PR #57, opened 2020-12-10
Update buildx ghaction

emerkle826:
This restores the ability to run the multi-platform docker buildx command with both `linux/amd64` and `linux/arm64` platform targets.


.github/workflows/ci.yaml
```diff
@@ -18,9 +18,11 @@ jobs:
           path: ~/.m2
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
+      - name: Set up QEMU
```
emerkle826, line 21: this is required to add the extra platforms (i.e. linux/arm64)

PR #59, opened 2020-12-10
Update for CASSANDRA-15299 protocol v5 changes

emerkle826:
This addresses the Cassandra 4.0 changes made as part of [CASSANDRA-15299](https://issues.apache.org/jira/browse/CASSANDRA-15299). Specifically, it changes the Netty channel pipeline initialization in the 4.x shim so that it works similar to how the Server changed for CASSANDRA-15299.


Dockerfile-oss
```diff
@@ -34,11 +46,24 @@ FROM oss311-${TARGETARCH} as oss311
...
+ENV CASSANDRA_PATH /opt/cassandra
+ENV MAAC_PATH /opt/management-api
+ENV MCAC_PATH /opt/metrics-collector
+
+ENV CASSANDRA_HOME ${CASSANDRA_PATH}
+ENV CASSANDRA_CONF ${CASSANDRA_PATH}/conf
+
+COPY --from=builder /build/management-api-common/target/datastax-mgmtapi-common-0.1.0-SNAPSHOT.jar ${MAAC_PATH}/
+COPY --from=builder /build/management-api-agent-3.x/target/datastax-mgmtapi-agent-3.x-0.1.0-SNAPSHOT.jar ${MAAC_PATH}/
+COPY --from=builder /build/management-api-agent-4.x/target/datastax-mgmtapi-agent-4.x-0.1.0-SNAPSHOT.jar ${MAAC_PATH}/
```
JeremiahDJordan, line 58: should we just only copy the jar that matches the C* version in the image?

management-api-server/pom.xml
```diff
@@ -183,7 +183,7 @@
                         <CASSANDRA_HOME>${basedir}/target/apache-cassandra-${cassandra.version}</CASSANDRA_HOME>
                     </systemPropertyVariables>
                     <environmentVariables>
-                        <JVM_EXTRA_OPTS>-javaagent:${basedir}/../management-api-agent/target/datastax-mgmtapi-agent-${project.version}.jar</JVM_EXTRA_OPTS>
+                        <JVM_EXTRA_OPTS>-javaagent:${basedir}/../management-api-agent/target/datastax-mgmtapi-agent-3.x-${project.version}.jar</JVM_EXTRA_OPTS>
```
JeremiahDJordan, line 186: Does this need a 4x version?

JeremiahDJordan (approved):
This looks good to me.  We should run this with cass-operator to double check everything there, since that is the main consumer, before merging.

management-api-server/pom.xml
emerkle826: As it stands now, it does not. This bit configures any unit tests that are run to have JVM_EXTRA_OPTS set, but none of the unit tests that run currently need this to be set to the 4.x agent. Most of the tests that are run are technically integration tests (though they run in the maven unit test phase), and spin up Docker containers with the appropriate image for the Cassandra/DSE version they are testing. The Docker images will only have the appropriate version of the agent jar in them in those tests.

tjake (changes requested):
Thanks for taking this on.

Overall, makes sense but there is a lot of redundant code now.  

I would suggest making an agent-common package with all the shared classes.  You could use the existing common module even though the server doesn't need it.

* There are also classes that aren't used like 4x in the 3x agent.  
* I don't think we need the shim loaders anymore.

Dockerfile-4_0
```diff
@@ -1,26 +1,63 @@
...
 RUN chmod +x /tini
 
 RUN set -eux; \
+  rm -fr /etc/apt/sources.list.d/*; \
+  rm -rf /var/lib/apt/lists/*; \
   apt-get update; \
   apt-get install -y --no-install-recommends wget iproute2; \
   rm -rf /var/lib/apt/lists/*
 
-ENV MCAC_VERSION 0.1.7
```
emerkle826, line 20: This needs to get added back in.

Dockerfile-astra-4_0
```diff
@@ -3,7 +3,8 @@ FROM management-api-for-apache-cassandra-builder as builder
 FROM datastax/astra:4.0
 
 COPY --from=builder /build/management-api-common/target/datastax-mgmtapi-common-0.1.0-SNAPSHOT.jar /tmp/
-COPY --from=builder /build/management-api-agent/target/datastax-mgmtapi-agent-0.1.0-SNAPSHOT.jar /tmp/
```
emerkle826, line 6: this needs to be updated like the 4.0 dockerfile

emerkle826:
@tjake I've refactored the agents so there are no more _shim_ projects. In doing so, I've introduced a couple of service loaders so we don't have to use reflection/class names directly in the code, each agent will load the appropriate providers for the bits that need to be different/compiled against different Cassandra libraries. If you have a few minutes for another review, I'd appreciate it.

JeremiahDJordan:
The refactoring looks good to me.  Thanks for the README about why some classes are duplicated.

tjake (approved):
LGTM thanks for cleaning that up!

PR #43, opened 2020-10-07
Run Cassandra 3.x and 4.x Docker images as cassandra

respringer:
This updates the Dockerfiles for Cassandra 3.x and 4.x to run as user "cassandra" and make "tini" the entrypoint.

PR #69, opened 2021-02-05
Run Cassandra 3.x and 4.x Docker images as cassandra

mapster:
Rebased @respringer 's #43 onto master, and added missing copy commands in entrypoint.


emerkle826:
@mapster Thanks for this, but I was already working on it in #59 which has been merged. The recently published v.0.1.20 images should run as `cassandra` now (3.11.7 - 3.11.10, as well as the 4.0 image that is based on 4.0-beta4) instead of using `gosu` to drop down.

PR #77, opened 2021-02-09
Revert USER change and use gosu

emerkle826:
This is a temporary change until k8ssandra permissions issues are resolved


emerkle826:
Holding on this PR as it might not be needed.

emerkle826:
This is no longer necessary

PR #80, opened 2021-02-12
Update to MCAC v0.2.0

emerkle826:
* Updates the MCAC agent jar to v0.2.0
* Allows for specifying CASSANDRA_LOGS to redirect startup logs

### DIFF
--- a/Dockerfile-4_0
+++ b/Dockerfile-4_0
@@ -2,7 +2,7 @@ ARG CASSANDRA_VERSION=4.0-beta4
 
 FROM maven:3.6.3-jdk-8-slim as builder
 
-ARG METRICS_COLLECTOR_VERSION=0.1.9
+ARG METRICS_COLLECTOR_VERSION=0.2.0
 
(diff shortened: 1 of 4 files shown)


JeremiahDJordan:
What’s the reason for changing the existing CASSANDRA_LOGS to a new location?  Will that not break existing logging sidecars that use the previous location?

JeremiahDJordan (approved):
+1 looks like the logs were already going to /var/log/cassandra so now the ENV var matches that.

emerkle826:
>  looks like the logs were already going to /var/log/cassandra so now the ENV var matches that.

That's correct. For OSS (3.11/4.0), the logs were going to `${CASSANDRA_PATH}/logs` which is effectively `/opt/cassandra/logs`, which is symlinked in the base Cassandra images to `/var/log/cassandra`

For DSE, the Dockerfile was effectively setting it to `/opt/dse/logs`, but the operational logs were still going to `/var/log/cassandra`.

Issue #86, opened 2021-02-18
high and critical CVEs io.netty package


zencircle:
Repository | Tag | CVE ID | Severity | Packages | Source Package | Package Version | Fix Status
-- | -- | -- | -- | -- | -- | -- | --
datastax/cassandra-mgmtapi-3_11_7 | v0.1.22 | CVE-2019-20445 | critical | io.netty_netty-all |   | 4.0.44.Final | fixed in 4.1.44
datastax/cassandra-mgmtapi-3_11_7 | v0.1.22 | CVE-2019-20444 | critical | io.netty_netty-all |   | 4.0.44.Final | fixed in 4.1.44
datastax/cassandra-mgmtapi-3_11_7 | v0.1.22 | CVE-2019-16869 | high | io.netty_netty-all |   | 4.0.44.Final | fixed in 4.1.42.Final

emerkle826:
This is another case, similar to #83 where the library in question is part of the Cassandra release itself, not the Management API. If the dependencies are updated in Cassandra and released,Management API will pick it up then.

zencircle:
https://issues.apache.org/jira/browse/CASSANDRA-16463

zencircle:
@emerkle826 Can you help me clear confusion Cassandra team does not think io.netty is part of Cassandra Package

emerkle826:
Looking at [this comment](https://issues.apache.org/jira/browse/CASSANDRA-16463?focusedCommentId=17289143&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17289143), it seems Cassandra does not make use of the affected Netty component, so the CVE is not applicable to Cassandra.

zencircle:
> Looking at [this comment](https://issues.apache.org/jira/browse/CASSANDRA-16463?focusedCommentId=17289143&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17289143), it seems Cassandra does not make use of the affected Netty component, so the CVE is not applicable to Cassandra.

In that case will these CVEs fall back on "cassandra-mgmtapi" ?, Ready to help and test a branch

emerkle826:
> In that case will these CVEs fall back on "cassandra-mgmtapi" ?, Ready to help and test a branch

[...]

Issue #85, opened 2021-02-18
critical CVE in logback package 


zencircle:
Tag | Distro | CVE ID | Severity | Packages | Source Package | Fix Package Version
-- | -- | -- | -- | -- | -- | --
v0.1.22 | Ubuntu-bionic | CVE-2017-5929 | critical | ch.qos.logback_logback-core | 1.1.3 | fixed in 1.2.0

emerkle826:
This is another case, similar to #83 where the library in question is part of the Cassandra release itself, not the Management API. If the dependencies are updated in Cassandra and released,Management API will pick it up then.

zencircle:
https://issues.apache.org/jira/browse/CASSANDRA-16464

Issue #83, opened 2021-02-16
high and critical CVE with resteasy libraries


zencircle:
Affected version Resteasy 4.0.0.Final
https://github.com/datastax/management-api-for-apache-cassandra/blob/master/management-api-server/pom.xml#L29
com.fasterxml.jackson.core_jackson-databind
CVE-2020-9548
CVE-2020-9547
CVE-2020-9546
CVE-2019-20330
CVE-2019-17267
CVE-2019-16335
CVE-2019-14540
CVE-2019-16943
CVE-2019-17531
CVE-2019-14893
CVE-2019-14892
CVE-2019-14379
CVE-2020-9548
CVE-2020-9547
CVE-2020-9546
CVE-2019-20330
CVE-2019-17267
CVE-2019-16335
CVE-2019-14540
CVE-2019-16943
CVE-2019-17531
CVE-2019-14893
CVE-2019-14892
CVE-2019-14379
Affected  2.9.9 fixed 2.9.10
More info
https://resteasy.github.io/index.html

emerkle826:
The new images have been published (v0.1.22 for each):
Cassandra 3.11.7: https://hub.docker.com/r/datastax/cassandra-mgmtapi-3_11_7/tags?page=1&ordering=last_updated
Cassandra 3.11.8: https://hub.docker.com/r/datastax/cassandra-mgmtapi-3_11_8/tags?page=1&ordering=last_updated
Cassandra 3.11.9: https://hub.docker.com/r/datastax/cassandra-mgmtapi-3_11_9/tags?page=1&ordering=last_updated
Cassandra 3.11.10: https://hub.docker.com/r/datastax/cassandra-mgmtapi-3_11_10/tags?page=1&ordering=last_updated

zencircle:
Thanks for fixes the criticals, but there at still 22 highs 


CVE ID | Severity | Packages | Source Package | Fixed Package Version
-- | -- | -- | -- | --
CVE-2020-24750 | high | com.fasterxml.jackson.core_jackson-databind | 2.9.10.4 | fixed in 2.9.10.6
CVE-2020-24616 | high | com.fasterxml.jackson.core_jackson-databind | 2.9.10.4 | fixed in 2.9.10.6
CVE-2020-14195 | high | com.fasterxml.jackson.core_jackson-databind | 2.9.10.4 | fixed in 2.9.10.5
[...]

Issue #78, opened 2021-02-09
Fix GID for cassandra


emerkle826:
In the 3.x and 4.x Dockerfiles, the Management API resources (as well as MCAC resources) are changed to be owned by the `cassandra` user, but have a group ID of `root`. The GID should be `cassandra`. For example
https://github.com/datastax/management-api-for-apache-cassandra/blob/master/Dockerfile-oss#L64

emerkle826:
Closing this as the group actually needs to be `root` when deploying with OpenShift

PR #87, opened 2021-02-19
Update dependencies

emerkle826:
Addresses #78, #83, #85, and #86


emerkle826:
@JeremiahDJordan or @tjake , I'd appreciate a review on this one. I'm essentially replacing jarfiles that ship with Cassandra 3.11.x with newer ones to pass some scans. I'm not worried about Logback or Jackson so much, but changing the version of Netty gives me some concern.

emerkle826:
Closing this as it should be addressed in Cassandra.

PR #89, opened 2021-03-01
Create home directory for cassandra user

### LINKED ISSUE #88: Fix cassandra user's HOME directory
See [this issue](https://github.com/k8ssandra/k8ssandra/issues/396)

Essentially, now that Management API runs as the `cassandra` user, that user's effective HOME directory should exist and be writable by that user. We can either:

1) Create the default HOME directory of `/home/cassandra` and set the correct permissions, or
2) Run a `usermod` command to alter the home directory to `/opt/cassandra` which is already setup with correct permissions.

careykevin:
Create the `/home/cassandra` directory and make it writable for the
`cassandra` user.

This will allow files, including the cqlsh history, to be written for
the user without warnings.

Resolves #88

### DIFF
--- a/Dockerfile-4_0
+++ b/Dockerfile-4_0
@@ -34,6 +34,7 @@ FROM cassandra:${CASSANDRA_VERSION}
 ENV CASSANDRA_PATH /opt/cassandra
 ENV MAAC_PATH /opt/management-api
 ENV MCAC_PATH /opt/metrics-collector
+ENV USER_HOME_PATH /home/cassandra
 
 ENV CASSANDRA_HOME ${CASSANDRA_PATH}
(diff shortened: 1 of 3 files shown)


emerkle826 (changes requested):
This looks good, but you need to make the same change to the DSE Dockerfile [here](https://github.com/datastax/management-api-for-apache-cassandra/blob/master/Dockerfile-dse-68)

emerkle826 (approved):

emerkle826:
@careykevin Thank you for the contribution. All I need you to do now is to sign our CLA policy here:
https://cla.datastax.com/
and ping me back on this PR when you have.

careykevin:
Thanks @emerkle826 - I have signed the agreement.

PR #92, opened 2021-03-09
dse-db support for api-agent-4.x 

jtgrabowski:


management-api-agent-4.x/pom.xml
```diff
@@ -59,16 +74,25 @@
...
+                    <groupId>com.datastax.dse</groupId>
+                    <artifactId>dse-db-all</artifactId>
+                    <version>${cassandra4.version}</version>
+                    <exclusions>
+                        <exclusion>
+                            <groupId>commons-codec</groupId>
+                            <artifactId>*</artifactId>
+                        </exclusion>
+                    </exclusions>
+                    <scope>provided</scope>
```
emerkle826, line 95: I assume this is needed because the DSE version you are using is Cassandra 4 based now?

There is already a DSE agent project that has DSE dependencies [here](https://github.com/datastax/management-api-for-apache-cassandra/blob/master/management-api-agent-dse-6.8/pom.xml#L50-L62).

Would it not be better to edit that POM file, or possibly create a new agent project that is for DSE with Cassandra 4 dependnecies?

jtgrabowski: Hi, we have a naming conflict here. 
The old `dse-db` jar comes from DSE. 
The new `dse-db-all` is this: https://github.com/datastax/cassandra. It's OSS C* 4.x. 
DSE agent would not be compatible with the new dse-db.

jtgrabowski: To clarify, datastax/cassandra may be a bit different than OSS C*. Now or in the future.

emerkle826: The problem I see is that the POM you are editing is responsible for building the 4.x agent that is released and used in the Docker images for Management API. What you added will still only build 1 4.x agent jarfile, but it will compile against either the Cassandra libs, or the Datastax version, based on the profile.

Are you only wanting the ability to build the 4.x agent against the Datastax version of the libs? As this PR stands, the 4.x agent won't get deployed or used in the Management API docker image compiled against the Datastax jars as the profile isn't enabled in the build/deploy/release steps.

If you want an automated build and deployment (to Maven) of the 4.x agent compiled against the Datastax libs, you'll need to either create a new agent sub module with a new name, or figure out how to edit the Shade plugin to create 2 shaded jars from the same project, with only the compile time dependencies changed. I wasn't able to come up with a way to do the latter when differentiating the 3.x and 4.x agents, so I had to create two sub modules.

emerkle826: If this is just for Datastax internal use, and not meant to be published like the rest of the artifacts, then it's fine. But if you want a 4.x agent to be published that happens to be compiled against these Datastax libs, you'll need to add the bits into the GitHub workflow files [here}](https://github.com/datastax/management-api-for-apache-cassandra/blob/master/.github/workflows/jar-release.yaml#L65) and [potentially here](https://github.com/datastax/management-api-for-apache-cassandra/blob/master/.github/workflows/ci.yaml#L66-L80) (for tests), assuming you end up creating a new sub project to bundle the new artifact.

jtgrabowski: Yes, I only want to be able to build against dse-db. 
I don't need an automated build/deployment.

management-api-server/pom.xml
```diff
@@ -2,14 +2,6 @@
...
     <modelVersion>4.0.0</modelVersion>
 
-    <repositories>
-        <repository>
-            <id>jcenter</id>
-            <name>jcenter</name>
-            <url>https://jcenter.bintray.com/</url>
-        </repository>
-    </repositories>
-
```
emerkle826, line 12: It looks like removing this repo is causing the CI build to fail.

https://github.com/datastax/management-api-for-apache-cassandra/actions/runs/636195617

```Could not find artifact com.aries:docker-java-shaded:jar:3.1.1 in central (https://repo.maven.apache.org/maven2)```

We either need to add it back, or adjust how the settings.xml file is constructed in the CI here:

https://github.com/datastax/management-api-for-apache-cassandra/blob/master/.github/workflows/ci.yaml#L48-L63

emerkle826 (changes requested):

management-api-server/pom.xml
emerkle826: Marking this resolved for now, but we'll have to address #93 at some point.

emerkle826 (approved):
As long as you don't need this commit to be tagged and released, I can merge it as is. The Maven cache issue with the tests still failing should be resolved in 7 days when the cache expires.

jtgrabowski:
thank you @emerkle826.
I don't need a tag nor release. 
I saw that all the tests passed on your temp branch. Just double checking (as I can't seam to find your temp branch anymore) did you simply restore the bintray repo without any other modifications? If yes, then let's merge!

emerkle826:
> thank you @emerkle826.
> I don't need a tag nor release.
> I saw that all the tests passed on your temp branch. Just double checking (as I can't seam to find your temp branch anymore) did you simply restore the bintray repo without any other modifications? If yes, then let's merge!

My branch is here:
https://github.com/emerkle826/management-api-for-apache-cassandra/tree/dse-db-support
and it's just a copy of yours and the build and tests seem to pass:
https://github.com/emerkle826/management-api-for-apache-cassandra/actions/runs/640096067

I'll go ahead and merge and keep an eye on the build stuff here.

Issue #101, opened 2021-04-22
DB home is not a writeable path


grighetto:
Running the API v0.1.24 on GKE fails with:

> Running java -Xms128m -Xmx128m -jar /opt/management-api/datastax-mgmtapi-server-0.1.0-SNAPSHOT.jar --cassandra-socket /tmp/cassandra.sock --host tcp://0.0.0.0:8080 --host file:///tmp/oss-mgmt.sock --explicit-start true --cassandra-home /var/lib/cassandra/
> Usage error: Option 'db_home' was given value '/var/lib/cassandra/' which is not a writeable path

Note: I'm running a slightly modified version of the Management API image because I need to run with a custom C* image instead of pulling the prebuilt one from dockerhub, but other than that my setup is pretty standard so I have no reasons to believe this wouldn't be a problem in general.

The issue is that when it mounts the volume /var/lib/cassandra in GKE, the folder ends up being owned by root (regardless of the `chown` that happened earlier). I confirmed that by logging the permissions at the very end of the Dockerfile execution which was correct, and then at the beginning of the entrypoint, which was incorrect. It looks like this is how k8s works: https://github.com/kubernetes/kubernetes/issues/2630

In the entrypoint, it does have another chown call to fix the permissions, but that only happens if the user is root: https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/v0.1.24/scripts/docker-entrypoint.sh#L17

I think we can fix this by removing/reverting the `USER cassandra` directive from the Dockerfile and then run the java command in the entrypoint with gosu.

grighetto:
I was able to fix this issue for me with https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/104

PR #104, opened 2021-04-28
#101 Fix "DB home is not a writeable path" on GKE

grighetto:


emerkle826:
Closing this for now as the issue seems to be resolved with cass-operator 1.7.0

PR #111, opened 2021-06-07
Fossa Integration

### LINKED ISSUE #109: Integrate Fossa component/license scanning
* Integrate via GH Actions to automate scanning and reporting
* Investigate how reports can be made public (hopefully via Fossa)

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-489) by [Unito](https://www.unito.io)

jdonenine:
Adds a GH action workflow that will automatically analyze the repo and updated public dependency report with all pushes to the master branch.

Fixed #109

┆Issue is synchronized with this [Jira Bug](https://k8ssandra.atlassian.net/browse/K8SSAND-547) by [Unito](https://www.unito.io)


.fossa.yml
```diff
@@ -0,0 +1,38 @@
...
+    target: pom.xml
+    path: management-api-agent-4.x
+  - name: datastax-mgmtapi-agent-common
+    type: mvn
+    target: pom.xml
+    path: management-api-agent-common
+  - name: datastax-mgmtapi-agent-dse-6.8
+    type: mvn
+    target: pom.xml
+    path: management-api-agent-dse-6.8
```
emerkle826, line 30: For the DSE agent, the project has to be built with the Maven profile `dse` enabled. I don't know i that has to be accounted for in the FOSSA config here or not. From the docs, it looks like it's going to run `mvn dependency:tree` to analyze things. If you don't do `mvn dependency:tree -Pdse`, it won't include the DSE agent. I'll look at the report to see if it actually is pickup up the DSE agent.

jdonenine: Gotcha, I'd have to do some digging or talk to Fossa to see if we can change the way it behaves for a certain module like that.  It certainly won't stop Fossa from scanning, so I think as it is it's OK, but we can see if we can make it a bit more accurate if necessary.

emerkle826 (approved):
This is fine for now. We can dig inot the DSE thing later if need be.

PR #117, opened 2021-07-16
Upgrade to 4.0.0 GA

jdonenine:
* Bump dependencies to the GA version for 4.0.0

### DIFF
--- a/management-api-agent-4.x/pom.xml
+++ b/management-api-agent-4.x/pom.xml
@@ -12,7 +12,7 @@
     <artifactId>datastax-mgmtapi-agent-4.x</artifactId>
 
     <properties>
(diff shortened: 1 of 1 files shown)


emerkle826:
@jdonenine We may want to also include 3.11.11 with this. It should just involve adding another section at the end of the docker release file [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/.github/workflows/docker-release.yaml#L136-L158).

According to the mailing list, 3.11.11 should be coming out soon.

jdonenine:
We'll just come back around in a few days to pickup the 3.11 patch as discussed.  4.0.0 has waited long enough 😄

PR #125, opened 2021-08-27
Add per dc replication factor overrides [K8SSAND-840]

### LINKED ISSUE #124: Replication of system keyspaces should be configurable per datacenter
The following system properties can be set to configure replication of system keyspaces:

* `cassandra.system_distributed_replication_dc_names`
* `cassandra.system_distributed_replication_per_dc`

When these properties are set the `SystemDistributedReplicationInterceptor` class configures replication for the following keyspaces:

* system_auth
* system_traces
* system_distributed

The `cassandra.system_distributed_replication_per_dc` configures the replication factor. The problem is that this value is used for all DCs. I have DCs with 1, 3, and 6 nodes for example, the replication factor will have to be set to 1.

Maybe we could introduce a new system property that accepts a comma-delimited list of key-value pairs which are separated by colons.  It might look something like this:

`cassandra.system_distributed_replication=dc1:1,dc2:3,dc3:3`

This would be easy to parse and provide the flexibility to configure the replication factor per DC.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-840) by [Unito](https://www.unito.io)
┆Fix Versions: management-api-for-apache-cassandra-0.1.29
┆Issue Number: K8SSAND-840
┆Priority: Medium

emerkle826:
Fixes #124

### DIFF
--- a/management-api-agent-common/src/main/java/com/datastax/mgmtapi/interceptors/SystemDistributedReplicationInterceptor.java
+++ b/management-api-agent-common/src/main/java/com/datastax/mgmtapi/interceptors/SystemDistributedReplicationInterceptor.java
@@ -15,6 +15,7 @@ import java.util.stream.Collectors;
 
 import com.google.common.collect.ImmutableMap;
 
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import net.bytebuddy.agent.builder.AgentBuilder;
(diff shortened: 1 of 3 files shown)


adejanovski (approved):
LGTM ✅ 
I was able to test the new override option with [this K8ssandra PR](https://github.com/k8ssandra/k8ssandra/pull/1093), adding an integration test.

PR #127, opened 2021-09-03
Add operation to list keyspaces

adejanovski:
Adds an operation to list existing keyspaces with a possibility to restrict the search to a single named keyspace.

### DIFF
--- a/management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
+++ b/management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
@@ -66,4 +66,8 @@ public interface CassandraAPI
     RpcStatementShim makeRpcStatement(String method, String[] params);
 
     HintsService getHintsService();
+
+    default List<String> getKeyspaces() {
+        return StorageService.instance.getKeyspaces();
+    }
 }
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 name: Java CI
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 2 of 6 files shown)


management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
```diff
@@ -66,4 +66,6 @@ default Object handleRpcResult(Callable<Object> rpcResult) throws Exception
     RpcStatementShim makeRpcStatement(String method, String[] params);
 
     HintsService getHintsService();
+
+    List<String> getKeyspaces();
```
emerkle826, line 70: Since the implementation for all 3 of the "shims" is the same, we could simple implement a `default` here in the interface, then the implementations wouldn't have to override anything.

emerkle826 (changes requested):
This looks good. The only change I'd request is to provide a `default` implementation in the CassandraAPI class, and remove the repeated implementations in the 3 shims.

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
adejanovski: Cool trick, I didn't know there was such a thing as a `default` implementation in Java interfaces 😅 
Always thought that's what abstract classes were for...
Thanks!

emerkle826: Yep, it's "new" as of Java 8. The use case for it is to be able to introduce new APIs without breaking older implementations that don't implement them.

emerkle826 (approved):
LGTM. :+1:

Issue #133, opened 2021-09-16
Review errors returned by management API


Miles-Garnsey:
It might prove profitable to do a full review of the errors returned by the management API. At present most errors returned tend to be 500s, with no error message included. 

For example, if keyspace creation fails because we've sent a bad request we get the following in the management API (or Cassandra?) logs:

```
Caused by: org.apache.cassandra.exceptions.ConfigurationException: Unrecognized strategy option {reaper-test} passed to NetworkTopologyStrategy for keyspace reaper_db
at org.apache.cassandra.locator.AbstractReplicationStrategy.validateExpectedOptions(AbstractReplicationStrategy.java:457)
at org.apache.cassandra.locator.NetworkTopologyStrategy.validateExpectedOptions(NetworkTopologyStrategy.java:303)
at org.apache.cassandra.locator.AbstractReplicationStrategy.validateReplicationStrategy(AbstractReplicationStrategy.java:402)
at org.apache.cassandra.schema.ReplicationParams.validate(ReplicationParams.java:78)
at org.apache.cassandra.schema.KeyspaceParams.validate(KeyspaceParams.java:94)
at org.apache.cassandra.cql3.statements.schema.CreateKeyspaceStatement.apply(CreateKeyspaceStatement.java:87)
at org.apache.cassandra.schema.Schema.transform(Schema.java:588)
at org.apache.cassandra.schema.MigrationManager.lambda$announce$2(MigrationManager.java:226)
at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
... 2 common frames omitted
```

But the client gets back this:

```
ERROR controllers.Reaper failed to create keyspace {"reaper": "default/reaper-test", "error": "incorrect status code of 500 when calling endpoint"} 
```

This seems not quite right, because we haven't encountered an internal error. We've sent the management API a bad request, which should register as a 400 error, not a 500. In addition, troubleshooting would be assisted by actually including the specifics of the error.

Miles-Garnsey:
Diving deeper into this codebase, I've found that a blocker for this issue might be the way that errors are communicated back from the agents to the management API server. 

[...]

Issue #128, opened 2021-09-08
K8SSAND-870 ⁃ IPCControllerTest fails when run on MacOS


emerkle826:
When trying to run IPCControllerTest on MacOS, the test fails with:

```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.447 s <<< FAILURE! - in com.datastax.mgmtapi.ipc.IPCControllerTest
[ERROR] clientServerTest(com.datastax.mgmtapi.ipc.IPCControllerTest)  Time elapsed: 0.414 s  <<< ERROR!
io.netty.channel.ChannelException: Unable to create Channel from class class io.netty.channel.epoll.EpollServerDomainSocketChannel
    at com.datastax.mgmtapi.ipc.IPCControllerTest.clientServerTest(IPCControllerTest.java:123)
Caused by: java.lang.reflect.InvocationTargetException
    at com.datastax.mgmtapi.ipc.IPCControllerTest.clientServerTest(IPCControllerTest.java:123)
Caused by: java.lang.NoClassDefFoundError: Could not initialize class io.netty.channel.epoll.Native
    at com.datastax.mgmtapi.ipc.IPCControllerTest.clientServerTest(IPCControllerTest.java:123)
```

Epoll is not available on MacOS, so it should try to use NIO in this case.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-870) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-870
┆Priority: Medium

emerkle826:
Little debugging on an old mac, I think I found the issue. When the IPCControllerTest starts up, it creates a netty server, which ends up initializing here:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-common/src/main/java/com/datastax/mgmtapi/ipc/IPCController.java#L75-L82
```java
        if (isClient)
        {
            b = b.channel(EpollDomainSocketChannel.class);
        }
        else
        {
            b = b.channel(EpollServerDomainSocketChannel.class);
        }
```
So the server bit in the test is forcing Epoll, which doesn't exist on MacOS

jsanda:
@emerkle826 nice find! Can you submit a PR to fix it?

PR #135, opened 2021-09-19
Miles/full query logging endpoint

Miles-Garnsey:
Draft PR. @emerkle826 I think this is wrong in a few ways but thought I'd get a draft out.

1. First up, I'm having some issues running the integration tests. Probably user error 😅
2. I also think I'm abusing `management-api-agent-common` by adding a 4x-only method to the `CassandraAPI.java`. I would have thought that this method should appear only in the 4x implementation `CassandraAPI4x`.

I'm inclined to think the RPC method `enableFullQuerylog()` [body](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/a47a4f8bb849a39523d13effe5fa1532fd5813b7/management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java#L70) should narrow the API instance by casting like this (solves problem 2 above by allowing us to remove the method from the common interface), but I don't think we can import the `CassandraAPI4x` type without creating a circular dependancy?

```
try {
            CassandraAPI cassAPI = (CassandraAPI4x) ShimLoader.instance.get();
            ShimLoader.instance.get().enableFullQuerylog();
            
        } catch (ClassCastException e ) {
            logger.error("Can't cast CassandraAPI instance to CassandraAPI4x. Full query logging is not supported on Cassandra < 4, are you sure you are running Cassandra 4.x +?");
            throw new UnsupportedOperationException()

        }
```

What are your thoughts on how to tackle this problem?

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ Finally build the Management API image:
 
 ### Standalone
 
+Note that the tests will not run correctly on MacOS because the IPC classes will fail.
+
     mvn -DskipTests package
     mvn test
     mvn integration-test -Drun311tests=true -Drun40tests=true
@@ -195,6 +197,18 @@ docker run -p 8080:8080 -it --rm mgmtapi-dse
     # Stop C*/DSE
     curl -XPOST --unix-socket /tmp/mgmtapi.sock http://localhost/api/v0/lifecycle/stop
     OK
+
+# Making changes
+
+The architecture of this repository is laid as follows, front-to-back:
+
+1. The `management-api-server/doc/openapi.json` documents the API.
+2. The server implements the HTTP verbs/endpoints under the `management-api-server/src/main/java/com/datastax/mgmtapi/resources` folder (e.g. `NodeOpsresources.java`).
+3. The server methods communicate back to the agents using `cqlService.executePreparedStatement()` calls which are routed as plaintext through a local socket. These calls return `ResultSet` objects, and to acess scalar values within these you are best to call `.one()` before checking for nulls and `.getObject(0)`. This java objectcan then be serialised into JSON for return to the client.
+4. The server communicates only with the `management-api-agent-common` subproject, which holds the unversioned `CassandraAPI` interface.
+5. The `management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java` routes commands through to specific versioned instances of `CassandraAPI` which is implemented in the version 3x/4x subprojects as `CassandraAPI4x`/`CassandraAPI3x`.
+
+Any change to add endpoints or features will need to make modifications in each of the above components to ensure that they propagate through.
 
 # CLI Help
   The CLI help covers the different options:
--- a/management-api-agent-4.x/src/main/java/com/datastax/mgmtapi/shim/CassandraAPI4x.java
+++ b/management-api-agent-4.x/src/main/java/com/datastax/mgmtapi/shim/CassandraAPI4x.java
@@ -36,6 +36,7 @@ import org.apache.cassandra.db.compaction.CompactionManager;
 import org.apache.cassandra.dht.IPartitioner;
 import org.apache.cassandra.dht.Range;
 import org.apache.cassandra.dht.Token;
+import org.apache.cassandra.fql.FullQueryLoggerOptions;
 import org.apache.cassandra.gms.ApplicationState;
 import org.apache.cassandra.gms.EndpointState;
 import org.apache.cassandra.gms.Gossiper;
@@ -66,7 +67,37 @@ public class CassandraAPI4x implements CassandraAPI
     private static final Logger logger = LoggerFactory.getLogger(CassandraAPI4x.class);
 
     private static final Supplier<SeedProvider> seedProvider = Suppliers.memoize(() -> new K8SeedProvider4x());
-
+    
+    @Override
+    public void enableFullQuerylog() 
+    {
+        logger.debug("Getting FQL options and calling enableFullQueryLogger.");
+        FullQueryLoggerOptions fqlOpts = DatabaseDescriptor.getFullQueryLogOptions();
+        StorageService.instance.enableFullQueryLogger(fqlOpts.log_dir, 
+            fqlOpts.roll_cycle, 
+            fqlOpts.block, 
+            fqlOpts.max_queue_weight, 
+            fqlOpts.max_log_size, 
+            fqlOpts.archive_command, 
+            fqlOpts.max_archive_retries);
+    }
+    
+    @Override
+    public void disableFullQuerylog() 
+    {
+        logger.debug("Stopping FullQueryLogger.");
+        StorageService.instance.stopFullQueryLogger();
+    }
+    
+    @Override
+    public boolean isFullQueryLogEnabled() 
+    {
+        boolean isEnabled = StorageService.instance.isFullQueryLogEnabled();
+        logger.debug("Querying whether full query logging is enabled. Result is {}", isEnabled);
+        return isEnabled;
+        
+    }
+    
     @Override
     public void decommission(boolean force) throws InterruptedException
     {
--- a/management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
+++ b/management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
@@ -66,6 +66,34 @@ public class NodeOpsProvider
         RpcRegistry.unregister(RPC_CLASS_NAME);
     }
 
+    @Rpc(name = "setFullQuerylog")
+    public void setFullQuerylog(@RpcParam(name="enabled") boolean fullQueryLoggingEnabled) throws UnsupportedOperationException
+    {
+        logger.debug("Attempting to set full query logging to " + fullQueryLoggingEnabled);
+        // Warning: bad design here. Default implementation will throw UnsupportedOperationException at runtime. Comparing strings to get versions is also ugly.
+        String cassVersion = ShimLoader.instance.get().getStorageService().getReleaseVersion();
+        if (Integer.parseInt(cassVersion.split("\\.")[0]) < 4){
+            logger.error("Full query logging is not available in Cassandra < 4x. This call is going to fail.");
+        }
+        if (fullQueryLoggingEnabled) {
+            ShimLoader.instance.get().enableFullQuerylog();
+        } else {
+            ShimLoader.instance.get().disableFullQuerylog();
+        }
+    }
+
+    @Rpc(name = "isFullQueryLogEnabled")
+    public boolean isFullQueryLogEnabled() throws UnsupportedOperationException
+    {
+        String cassVersion = ShimLoader.instance.get().getStorageService().getReleaseVersion();
+        logger.debug("Attempting to retrieve full query logging status for cassandra version" + cassVersion + ".");
+        if (Integer.parseInt(cassVersion.split("\\.")[0]) < 4){
+            logger.error("Full query logging is not available in Cassandra < 4x. This call is going to fail.");
+        }
+        logger.debug("Calling ShimLoader.instance.get().isFullQueryLogEnabled()");
+        return ShimLoader.instance.get().isFullQueryLogEnabled();
+    }
+
     @Rpc(name = "reloadSeeds")
     public List<String> reloadSeeds()
     {
--- a/management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
+++ b/management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
@@ -29,6 +29,23 @@ import org.apache.cassandra.transport.Server;
  */
 public interface CassandraAPI
 {
+    
+    default public void enableFullQuerylog()
+    {
+        throw new UnsupportedOperationException("FQL is only supported on OSS Cassandra > 4x.");
+    }
+    
+    default public void disableFullQuerylog() 
+    {
+        throw new UnsupportedOperationException("FQL is only supported on OSS Cassandra > 4x.");
+    }
+    
+    default public boolean isFullQueryLogEnabled() 
+    {
+        throw new UnsupportedOperationException("FQL is only supported on OSS Cassandra > 4x.");
+        
+    }
+
     void decommission(boolean force) throws InterruptedException;
 
     Map<List<Long>, List<String>> checkConsistencyLevel(String consistencyLevelName, Integer rfPerDc);
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ local.properties
 *.ipr
 out/
 .idea_modules/
+.vscode/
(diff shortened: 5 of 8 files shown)


emerkle826:
>     1. First up, I'm having some issues running the integration tests. Probably user error

If you are running the integration tests locally on MacOS, there is a known issue where one of the tests incorrectly tries to load Netty's epoll stuff, which doesn't exist for MacOS (issue #128).

The other issue with the integration tests is DSE tests will fail in GitHub Actions if your fork of the repo doesn't have GitHub secrets for authenticating to the DSE artifactory.

emerkle826:
>     2. I also think I'm abusing `management-api-agent-common` by adding a 4x-only method to the `CassandraAPI.java`. I would have thought that this method should appear only in the 4x implementation `CassandraAPI4x`

So the way the Agent side of the project is designed, we'll have to add the method to the common API, even though it's only applicable to C* 4.0+. I'll add some more detailed comments in the PR itself.

emerkle826:
I'm still debating the implementation details on this. On one hand, if the Agent throws an exception when the operation is not supported, then attempting it on 3.11 will get an error all the way back to the server, which will log something useful in the Management API logs as to why it failed. However, the REST server side of Management API is the only consumer of the RPC/Agent at the moment. Throwing an exception from the Agent back to the server seems a bit clunky.

The other way to go about it would be to do what I believe you tried in an earlier iteration and have the REST server side of the code actually cut off the request if it detected that Cassnadra isn't 4.0 or greater. I tend to prefer that approach as the RPC call doesn't even make it to the agent. The problem is that this couples the logic to the REST server. If we were to ever support some other agent RPC consumer, it would be on that new consumer to handle this case as the Agent wouldn't contain the logic.

A compromise between the two approaches could be to let the 3.11 agent implementation just be a no-op (with a log perhaps), as opposed to throwing an Exception, and the 4.0 agent would actually manipulate the query logging options. The drawback here is that any 3.11 logging would be in just the agent (i.e. embedded with the actual Cassandra logs) and it might not be as obvious that the operation isn't doing anything, as the REST server side would just return the same 200 OK (same response if the operation is successful against a 4.0+ Cassandra).

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
```diff
@@ -30,6 +30,12 @@
 public interface CassandraAPI
 {
     void decommission(boolean force) throws InterruptedException;
+    
+    void enableFullQuerylog() throws UnsupportedOperationException;
+
+    void disableFullQuerylog()throws UnsupportedOperationException;
```
emerkle826, line 36: I don't have a strong opinion here, but you could condense these two into a `setFullQueryLog(boolean enabled)` API, like you did with the actual REST endpoint. Either way is fine though.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -66,6 +66,23 @@ public synchronized void unregister()
         RpcRegistry.unregister(RPC_CLASS_NAME);
     }
 
+    @Rpc(name = "setFullQuerylog")
+    public void setFullQuerylog(@RpcParam(name="enabled") boolean fullQueryLoggingEnabled) throws UnsupportedOperationException
+    {
+        // Warning: bad design here. Implementations for 3.x adn DSE will throw UnsupportedOperationException at runtime.
```
emerkle826, line 72: Again, not sure which way is best. Let the 3.11 implementation throw an Exception, or check the version here to cut off the call before it even goes to the agent specific shim.

management-api-agent-4.x/src/main/java/com/datastax/mgmtapi/shim/CassandraAPI4x.java
```diff
@@ -67,6 +68,29 @@
 
     private static final Supplier<SeedProvider> seedProvider = Suppliers.memoize(() -> new K8SeedProvider4x());
 
+    public void enableFullQuerylog() 
+    {
+        FullQueryLoggerOptions fqlOpts = DatabaseDescriptor.getFullQueryLogOptions();
```
emerkle826, line 73: When I briefly looked at this part, I wasn't sure if you had to give the service all the options or not. Do you know if there is a simple `enable` function that will just use whatever FQL config Cassandra has in its `cassandra.yaml`? I so, this can be simplified as there is no need to grab the config it already has and pass to the service call if the service will just use the same config be default. I think making this call here would be more for if the Management API REST endpoint allowed for configuration overrides, which I don't think we want to do at this point.

emerkle826 (changes requested):
Looks good so far, though I think we want to discuss a little more about the implementation details.

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
emerkle826: Also, you will have to implement these API methods in the DSEE agent [here](https://github.com/Miles-Garnsey/management-api-for-apache-cassandra/blob/miles/full-query-logging-endpoint/management-api-agent-dse-6.8/src/main/java/com/datastax/mgmtapi/shim/DseAPI68.java#L66)

By default, it won't try to build the DSE agent (as you have to have access to the Datasatx DSE Maven artifactory repo to get the DSE jars). To build the DSE agent, you have to enable the DSE Maven profile, so something like:

```
mvn clean verify -P dse -DrunDSEtests
```

Miles-Garnsey:
> I'm still debating the implementation details on this. On one hand, if the Agent throws an exception when the operation is not supported, then attempting it on 3.11 will get an error all the way back to the server, which will log something useful in the Management API logs as to why it failed. However, the REST server side of Management API is the only consumer of the RPC/Agent at the moment. Throwing an exception from the Agent back to the server seems a bit clunky.

Thanks for the comments @emerkle826 you've summarised the agonising I've done over this design very succinctly! Ideally we'd like to throw all the way back to the REST client, but I've written some notes on why this may be hard (#133).

> The other way to go about it would be to do what I believe you tried in an earlier iteration and have the REST server side of the code actually cut off the request if it detected that Cassnadra isn't 4.0 or greater. I tend to prefer that approach as the RPC call doesn't even make it to the agent. The problem is that this couples the logic to the REST server.

What are the implications of throwing an error in the agent? Is it open to us to (1) log an error from the server and (2) also throw an error in the agent for 3x? Or will that be fatal to the agent? 

>  If we were to ever support some other agent RPC consumer, it would be on that new consumer to handle this case as the Agent wouldn't contain the logic.

Yes, this was a bugbear for me too. In fairness if a new agent is implementing the `CassandraAPI` interface I guess that this is explicit to the person implementing.

> A compromise between the two approaches could be to let the 3.11 agent implementation just be a no-op (with a log perhaps), as opposed to throwing an Exception, and the 4.0 agent would actually manipulate the query logging options. The drawback here is that any 3.11 logging would be in just the agent (i.e. embedded with the actual Cassandra logs) and it might not be as obvious that the operation isn't doing anything, as the REST server side would just return the same 200 OK (same response if the operation is successful against a 4.0+ Cassandra).

I don't love the idea of returning a 200 for an invalid request 😞 Can we maybe do a hybrid approach where we do a version comparison based on strings (yes, horrible) in the server, log an error (but proceed anyway and throw no error there), then implement in `CassandraAPI4x` and throw an `UnsupportedOperationException` in `CassandraAPI3x`?

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
Miles-Garnsey: DSE part implemented. 

I also thought about where we should split things off into multiple enable/disable functions. I don't think there's a right answer, so I'm happy doing it this way.

management-api-agent-4.x/src/main/java/com/datastax/mgmtapi/shim/CassandraAPI4x.java
Miles-Garnsey: I think I do. I don't believe there is a method to enable FQL that doesn't take the parms (happy to be wrong though!)

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
Miles-Garnsey: Have a look at the changes I've made. log an error in the server, throw the error in the 3x implementation.

Miles-Garnsey:
I think this is good to go. I'm not certain on the DSE side of things, but that is now implemented as 'correctly' as I can make it. Not sure how to test that this compiles though - I may have to look into getting Artefactory access etc.

Tested on both 4.0.0 and 3.11.10 clusters. The `.../fullquerylogging` endpoint fails on 3x clusters with a 500 status but it doesn't bring the cluster down. We do see the logging I want to see (in Cassandra's logs rather than in the server where we'd probably prefer them, but anyway...)

README.md
```diff
@@ -196,6 +198,18 @@ docker run -p 8080:8080 -it --rm mgmtapi-dse
...
+The architecture of this repository is laid as follows, front-to-back:
+
+1. The `management-api-server/doc/openapi.json` documents the API.
+2. The server implements the HTTP verbs/endpoints under the `management-api-server/src/main/java/com/datastax/mgmtapi/resources` folder (e.g. `NodeOpsresources.java`).
+3. The server methods communicate back to the agents using `cqlService.executePreparedStatement()` calls which are routed as plaintext through a local socket. These calls return `Object` which is then serialised into JSON for return to the client.
+4. The server communicates only with the `management-api-agent-common` subproject, which holds the unversioned `CassandraAPI` interface.
+5. The `management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java` routes commands through to specific versioned instances of `CassandraAPI` which is implemented in the version 3x/4x subprojects as `CassandraAPI4x`/`CassandraAPI3x`.
+
+Any change to add endpoints or features will need to make modifications in each of the above components to ensure that they propagate through.
+
```
emerkle826, line 212: Thanks for documenting this. It's all stuff you will learn the hard way when you work on it, but makes it much easier to start when it's laid out for you. Awesome!

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -66,6 +66,34 @@ public synchronized void unregister()
...
 
+    @Rpc(name = "setFullQuerylog")
+    public void setFullQuerylog(@RpcParam(name="enabled") boolean fullQueryLoggingEnabled) throws UnsupportedOperationException
+    {
+        logger.debug("Attempting to set full query logging to " + fullQueryLoggingEnabled);
+        String cassVersion = ShimLoader.instance.get().getStorageService().getReleaseVersion();
+        if (Integer.parseInt(cassVersion.split("\\.")[0]) < 4){
+            logger.error("Full query logging is not available in Cassandra < 4x. This call is going to fail.");
+        }
+        // Warning: bad design here. Implementations for 3.x and DSE will throw UnsupportedOperationException at runtime.
```
emerkle826, line 77: nit: You can remove this comment now. Or maybe move it p above the Cassandra 4.0+ check above?

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
```diff
@@ -30,6 +30,12 @@
...
+    
+    boolean isFullQueryLogEnabled() throws UnsupportedOperationException;
```
emerkle826, line 38: So for these 3 new methods, I think a better approach would be to provide a default implementation in the interface here, and only override them in CassandraAPI4x.
```suggestion
    default void enableFullQuerylog() throws UnsupportedOperationException
    {
        throw new UnsupportedOperationException("FQL is only supported on OSS Cassandra > 4x");
    };

    default void disableFullQuerylog()throws UnsupportedOperationException;
    {
        throw new UnsupportedOperationException("FQL is only supported on OSS Cassandra > 4x");
    };

    default boolean isFullQueryLogEnabled() throws UnsupportedOperationException;
    {
        throw new UnsupportedOperationException("FQL is only supported on OSS Cassandra > 4x");
    };
```

With the above, you can remove the overrides in the CassandraAPI3x class and the DseAPI68 class.

management-api-agent-4.x/src/main/java/com/datastax/mgmtapi/shim/CassandraAPI4x.java
```diff
@@ -67,6 +68,33 @@
 
     private static final Supplier<SeedProvider> seedProvider = Suppliers.memoize(() -> new K8SeedProvider4x());
 
+    public void enableFullQuerylog()
```
emerkle826, line 71: nit: Add the `@Override` annotation on these three methods. Just a style thing :wink:

emerkle826 (changes requested):
This looks good. Just a few more small changes and it's good to go.

README.md
Miles-Garnsey: You're welcome! Glad you like it as I wasn't sure if I'd gotten it 100% right!

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
Miles-Garnsey: I like the direction of this thought... So the idea is we default to `UnsupportedOperationException("FQL is only supported on OSS Cassandra > 4x");` and override for 4x. 

My only question: we may want to switch this around so that the override is on the CassandraAPI3x class and the DseAPI68 class. This is because the feature will exist in hypothetical future CassandraAPI5x classes etc. so eventually it is the error case that will be the exception.

management-api-agent-4.x/src/main/java/com/datastax/mgmtapi/shim/CassandraAPI4x.java
Miles-Garnsey: Fair. Will do.

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
Miles-Garnsey: Oh also, these will be the only methods with default implementations in that interface. Just noting in case we want to reconsider for consistency.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
Miles-Garnsey: Done.

Miles-Garnsey:
Squashed, ready for merge (but I haven't been able to test DSE yet because I need Artifactory access). For some reason I still can't see a merge button here - I see `Only those with write access to this repository can merge pull requests.` Grrr.

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
emerkle826: Yes, the `default` mechanism here is meant to provide the most basic functionality across all implementations and can't access anything not at the interface level. Since we have 2 agents that will just thrown an unsupported Exception (3.11 and DSE 6.8), that's about all we can do here.
Going forward, a Cassandra 5.x agent might need to do something different than the 4.x code, so there's no guarantees that a common implementation would exist between them (however unlikely that is in this specific case).

emerkle826:
> Squashed, ready for merge (but I haven't been able to test DSE yet because I need Artifactory access). For some reason I still can't see a merge button here - I see `Only those with write access to this repository can merge pull requests.` Grrr.

I have given you access to merge. I'm running the DSE tests locally. Let me know if you have trouble getting Artifactory access from the link I sent you and I'll do some more digging.

emerkle826 (approved):
LGTM. When this is merged in to master, GitHub will run the DSE tests and they should pass.

PR #140, opened 2021-09-28
Non-root verification / bump Tini

jeffbanks:
Reviewed as part of k8ssandra security [epic](https://k8ssandra.atlassian.net/browse/K8SSAND-705).

- Related to [this effort](https://k8ssandra.atlassian.net/browse/K8SSAND-802) as part of security hardening for running entrypoint script as non-root.
- The primary mgmt-api images used by k8ssandra at current were examined with an update to Tini's version where needed.
- Two images were using `cassandra` user and one using `dse` user.


emerkle826 (approved):
This looks good to me.

PR #141, opened 2021-10-05
Execute forceKeyspaceCleanup in the background

burmanm:
Since cleanup process can take a very long time, we should not force caller to wait for the execution to finish. Instead, execute it in the background.

The user is able to verify the status of the cleanup job in >= 4.0 by checking the status of SSTable jobs from the virtual table system.sstable_tasks

I selected a single background thread to avoid overflowing the Cassandra itself with too many requests.


burmanm:
.. or should I target this to 4.0-management-api only?

emerkle826 (changes requested):
This looks ok to me. The only issue is that the client calling this operation now gets an immediate response, and it's likely always a 200 OK (unless launching the background task fails). Checking the success of the operation is now the responsibility of the client making the call (if success matters to the client). Failure detection is also the responsibility of the client.

Could you edit the description here:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-server/doc/openapi.json#L217
just to indicate that the operation is async now?

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -203,16 +209,25 @@ public void forceKeyspaceCleanup(@RpcParam(name="jobs") int jobs,
...
+        } else {
+            keyspaces.add(keyspaceName);
         }
 
-        for (String keyspace : keyspaces)
-        {
-            ShimLoader.instance.get().getStorageService().forceKeyspaceCleanup(jobs, keyspace, tables.toArray(new String[]{}));
-        }
+        // Send to background execution
+        service.execute(() -> {
```
jsanda, line 222: How do we track the progress on the client side?

jsanda:
Would it be worthwhile to keep a sync version as well?

emerkle826:
> Would it be worthwhile to keep a sync version as well?

If there is no way to track the progress, maybe we should.

jsanda:
>Also, I assume this job map doesn't need to survive restarts, correct? If so, then there's no need to serialize the map to disk and deserialize it after a restart.

Correct. No need to survive restarts.

jeffbanks:
This all sounds good, like that the rest endpoint waiting will be removed.  What retry period are you thinking for the client in this case?

PR #143, opened 2021-10-05
More endpoints: create table, alter table, list tables, get keyspace replication

adutra:
Done so far:

- [x] Create table
- [ ] Alter table
- [x] List tables
- [x] Get keyspace replication


adutra:
Thanks for the heads up @emerkle826 !

This is now ready for review. I'm leaving out the alter table endpoint as it requires a lot more work, and we don't  need it right now.

PR #155, opened 2021-10-18
Add full_query_logging to featureSet

Miles-Garnsey:
Supports [cass-operator 176](https://github.com/k8ssandra/cass-operator/pull/176) using the new featureset endpoint.


Miles-Garnsey:
@burmanm and @emerkle826 - are you OK with this (extremely minor) change? It supports my cass-operator PR by allowing us to query for whether the FQL endpoints are available.

PR #162, opened 2021-11-10
Fix NPE in CallStatus and change return code to 404

burmanm:
If the job did not exists (the pod crashed while executing the job), then management-api would return NPE and 500. This PR changes the behaviour to return 404 in that case indicating the job_id did not match any known jobs to the jobExecutor.


emerkle826 (approved):

emerkle826:
@burmanm You can merge this (rebase and merge please) if you are done with it.

PR #172, opened 2022-01-13
add endpoint to list schema versions, needed to check schema agreement

jsanda:
We can already get schema versions from the `/metadata/endpoints` endpoint, but that's not entirely intuitive. I am adding an `/ops/node/schema/versions` endpoint which should be a bit simpler. It just delegates to `StorageProxy.describeSchemaVersions()` which is what `nodetool describecluster` does as well.

### DIFF
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/resources/v1/NodeOpsResources.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/resources/v1/NodeOpsResources.java
@@ -74,7 +74,12 @@ public class NodeOpsResources {
     @GET
     @Path("/schema/versions")
     @Produces(MediaType.APPLICATION_JSON)
-    @ApiResponse(responseCode = "200", description = "Gets the schema versions for each node. Useful for checking schema agreement")
+    @ApiResponse(
+            responseCode = "200",
+            description = "Gets the schema versions for each node. Useful for checking schema agreement",
+            content = @Content(
+                    mediaType = MediaType.APPLICATION_JSON,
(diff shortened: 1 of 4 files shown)


management-api-server/src/main/java/com/datastax/mgmtapi/resources/v1/NodeOpsResources.java
```diff
@@ -59,10 +65,30 @@ public Response decommission(@QueryParam(value="force")boolean force)
...
+        return handle(() ->
                 Response.accepted(
                         ResponseTools.getSingleRowStringResponse(app.dbUnixSocketFile, cqlService, "CALL NodeOps.rebuild(?)", srcDatacenter)
                 ).build());
     }
 
+    @GET
+    @Path("/schema/versions")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiResponse(responseCode = "200", description = "Gets the schema versions for each node. Useful for checking schema agreement")
```
emerkle826, line 77: To make the OpenAPI spec generation accurate, you need to add an example content to `APIResponse`. It should probably be like this:
```java
  @ApiResponse(
      responseCode = "200",
      description = "Gets the schema versions for each node. Useful for checking schema agreement",
      content =
          @Content(
              examples = @ExampleObject(value = "{2207c2a9-f598-3971-986b-2926e09e239d: [10.244.1.4, 10.244.2.3, 10.244.3.3]}")))
```

Once you have this, a `mvn compile` should update the openapi spec file too

emerkle826 (changes requested):
Just need to update the APIResponse annotation with a response example and it should be good.

emerkle826 (approved):

PR #185, opened 2022-04-21
Include CDC agent in images

### LINKED ISSUE #184: K8SSAND-1425 ⁃ Include CDC agent in management API image
The Datastax CDC agent needs to be included in the management API images for both OSS Apache Cassandra and DSE. 

We probably don’t want to enable the agent unless the user requests CDC be enabled (for performance reasons) but given it is controlled via `ADDITIONAL_JVM_OPTS` we shouldn’t need to add anything specific to the images to facilitate that.

My preliminary understanding is that the agent runs in the same JVM as Cassandra, and as a result the agent cannot be run in a separate container.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1425) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://k8ssandra.atlassian.net/rest/api/2/attachment/content/10004">Screen Shot 2022-04-19 at 6.49.50 PM.png</a>
┆friendlyId: K8SSAND-1425
┆priority: Medium

Miles-Garnsey:
All images updated to include DataStax CDC agent for Apache Cassandra or DSE (as applicable).

* Apache Cassandra 3.0.x, 4.0.x images updated to include CDC agent for Apache Cassandra.
* DSE images updated to include CDC agent for DSE.

**Issue addressed**

Closes #184.


Miles-Garnsey:
Do not merge yet. I want to test this from cass-operator in case I need to make additional changes.

emerkle826:
@miles I'm not sure how CDC is enabled exactly, but if it's an Agent jarfile, does it need to be added to the Cassandra classpath on startup? Similar to how the Management API jarfile is added here:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/scripts/docker-entrypoint.sh#L84-L89

If so, a similar change would be needed here:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/scripts/dse-6.8-docker-entrypoint.sh#L45-L50
for DSE images. Although, the DSE images produced here are not used by cass-operator (yet).

emerkle826:
Maybe we don't need to update the entrypoint scripts. According to the [docs](https://docs.datastax.com/en/cdc-for-cassandra/cdc-apache-cassandra/1.0.2/install.html), all we need to do is set `JVM_EXTRA_OPTS` to specify the agent/path, which will be done by k8ssandra/cass-operator. So maybe this PR is fine as is.

Miles-Garnsey:
> Maybe we don't need to update the entrypoint scripts. According to the [docs](https://docs.datastax.com/en/cdc-for-cassandra/cdc-apache-cassandra/1.0.2/install.html), all we need to do is set `JVM_EXTRA_OPTS` to specify the agent/path, which will be done by k8ssandra/cass-operator. So maybe this PR is fine as is.

That's right. No changes required to the classpath here, this is a Java agent and enabled by JVM_EXTRA_OPTS.

I'm more concerned about [these](https://github.com/k8ssandra/management-api-for-apache-cassandra/runs/6106834424?check_suite_focus=true) failing tests, do you know what would cause this as I don't think it should have anything to do with this PR.

emerkle826:
Try running the 3.11 tests locally. If you don't reproduce the problem locally, try re-running the GHA test. If it passes on the re-run, I'd say we have a flaky test or possibly some race condition. If the test fails again with the same errors, I'll try to help dive in.

jsanda:
How much does this increase image sizes?

Miles-Garnsey:
> How much does this increase image sizes?

40MB at this stage.

emerkle826 (approved):

emerkle826:
@Miles-Garnsey If you don't mind, I think pulling in the changes I have [here](https://github.com/Miles-Garnsey/management-api-for-apache-cassandra/pull/5) would be huge as it would cut down the Docker image size by almost 170MB from their current sizes, and doesn't add any noticeable size with your CDC agent additions.

PR #200, opened 2022-07-05
Use same path for CDC agent irrespective of C*/DSE version/agent version

Miles-Garnsey:
Moving CDC agent to /opt/cdc_agent/cdc-agent.jar for all agent versions and all C*/DSE versions. As requested by @emerkle826 and @burmanm.


emerkle826:
@Miles-Garnsey let me know if you have trouble merging and/or releasing. All that you need to do to release is push a new tag once the change is merged into master.

PR #208, opened 2022-09-22
Updaet DSE image builds based on docker-images

emerkle826:
Updates the DSE image builds to match (as closely as possible) the official builds for DSE Docker images with Management API embedded.

Fixes: #209


.github/workflows/docker-release.yaml
```diff
@@ -10,9 +10,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        cassandra-version: [6.8.25]
+        dse-version: [6.8.25, 6.8.26]
```
emerkle826, line 13: I'm not sure if we want/need to go back to 6.8.25 or not. The list here should only contain DSE versions we will support going froward. Also, since DSE images built by this action are only for testing, maybe we don't even need to maintain old versions here?

dse-68/Dockerfile.jdk11
```diff
@@ -0,0 +1,180 @@
+# Copyright DataStax, Inc, 2017
```
emerkle826, line 1: All of the Docker build files were generated from the `docker-images` repo with the changes from this PR:
https://github.com/riptano/docker-images/pull/87
There are some slight modifications to the build files as that PR will add a released Management API bundle to the DSE image, where the files in this PR will compile the latest Management API bundles and add them to the images built in this repo.

.github/workflows/docker-release.yaml
adejanovski: not sure either... We may want to leave it like this for now though.

.github/workflows/ci.yaml
```diff
@@ -131,6 +131,9 @@ jobs:
     if: ${{ github.ref == 'refs/heads/master' && github.event_name == 'push'}}
     needs: build-dse
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        jdk-version: [jdk8, jdk11]
```
adejanovski, line 136: Do we need the jdk8 builds?

adejanovski (approved):
Looks good, pending CI passing and probably a line that should be uncommented before merging?

.github/workflows/ci.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
         run: |
           mvn -B -q install --file pom.xml -Drun311tests=${{ matrix.run311tests }} -Drun40tests=${{ matrix.run40tests }}
   build-dse:
-    if: ${{ github.ref == 'refs/heads/master' && github.event_name == 'push'}}
+#    if: ${{ github.ref == 'refs/heads/master' && github.event_name == 'push'}}
```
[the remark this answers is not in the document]
emerkle826: Yes, that was commented out to get the DSE tests to run. They normally don't run until things are merged to master as this was the best I could do to prevent DSE tests from running on PRs from forks.

.github/workflows/ci.yaml
emerkle826: We don't necessarily, but I believe the cass-operator default DSE image is a JDK8 based image. The official DSE images have a `-java11` suffix to the image tag for JDK11 based images, while the JDK8 based images do not have a suffix for the tag.

emerkle826: Also, the DSE images aren't official, just used for testing at the moment. But it doesn't hurt to build both as the official build process will.

.github/workflows/docker-release.yaml
emerkle826: No real harm in doing so.

PR #227, opened 2022-11-16
Add GH Projects automation

adejanovski:
Add automation for GH projects:

- new issues get added to the board automatically
- PRs are checked for a linked issue (non blocking, just shows up in the check list)

### DIFF
--- a/.github/workflows/add_to_project.yaml
+++ b/.github/workflows/add_to_project.yaml
@@ -1,0 +1,16 @@
+name: Add issues to GH project
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  add-to-project:
+    name: Add issue to GH project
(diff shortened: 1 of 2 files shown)


.github/workflows/add_to_project.yaml
```diff
@@ -0,0 +1,16 @@
+name: Add issues to GH project
```
adutra, line 1: Don't we have license headers in place for this project?

adutra (approved):

.github/workflows/add_to_project.yaml
adejanovski: It doesn't seem to be the case, looking at the other workflows.
The `license-check` workflow file is about Fossa scanning, which is slightly different.

PR #236, opened 2022-12-14
Update ByteBuddy to 1.12.19

burmanm:
Updates ByteBuddy to 1.12.19. Started fine at least with Cassandra 4.0.6

### DIFF
--- a/management-api-agent-common/src/main/java/com/datastax/mgmtapi/interceptors/CassandraDaemonInterceptor.java
+++ b/management-api-agent-common/src/main/java/com/datastax/mgmtapi/interceptors/CassandraDaemonInterceptor.java
@@ -5,31 +5,28 @@
  */
 package com.datastax.mgmtapi.interceptors;
 
-import java.io.File;
-import java.nio.file.Paths;
-import java.util.concurrent.Callable;
-
-import com.google.common.collect.ImmutableMap;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
+import com.datastax.mgmtapi.NodeOpsProvider;
 import com.datastax.mgmtapi.ShimLoader;
 import com.datastax.mgmtapi.ipc.IPCController;
-import com.datastax.mgmtapi.NodeOpsProvider;
+import com.google.common.collect.ImmutableMap;
 import io.netty.channel.ChannelOption;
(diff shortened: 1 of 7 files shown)


emerkle826:
@burmanm Looks like the integration tests mostly failed. Can you try to restart them to see if they start to pass? Briefly looking at the logs, it looks like the test timed out waiting for Cassandra to become ready, but the Cassandra log dumps seem to show Cassandra is ready, however the Management API socket may not be. Not sure. Were you able to run any of the tests locally?

emerkle826:
You can run a specific integration test like this:
```sh
mvn verify -Dit.test=LifecycleIT -DfailIfNoTests=false -Drun40tests
```

burmanm:
None of the tests work on my machine, so I can't say. All I get is spam of [WARNING] We have a duplicate com/datastax/dse/protocol/internal/DseProtocolConstants$Opcode.class

etc

And then a failure of random something before tests even really start.

burmanm:
Okay, that was solved (partly - there's still maven warnings) updating all the super-old maven-shade plugin versions.

emerkle826:
I can try to run one locally. It may just be that the tests in general need some work with this update, but the images are fine.

burmanm:
But I have no idea why we have these:

```
Status 409: {"message":"Conflict. The container name \"/mgmtapi\" is already in use by container \"3dfe7a1a042e145ea036602d128b203b21abd1102a31dd9cf351e20c975fba7a\". You have to remove (or rename) that container to be able to reuse that name."}
```

emerkle826:
> But I have no idea why we have these:
> 
> ```
> Status 409: {"message":"Conflict. The container name \"/mgmtapi\" is already in use by container \"3dfe7a1a042e145ea036602d128b203b21abd1102a31dd9cf351e20c975fba7a\". You have to remove (or rename) that container to be able to reuse that name."}
> ```

That is likely due to running the test locally and them not completing successfully. The test will try to build a Docker image and then run a container named "mgmtapi" using the built image. The tests should clean up after themselves, but I guess there might be some cases where the cleanup code doesn't run and the container remains afterwards.

If you manually delete the container, you shouldn't have that error again unless you run into the same situation.

burmanm:
#238 solves this also..

PR #238, opened 2022-12-15
New metrics collector implementation

### LINKED ISSUE #228: Expose the Cassandra metrics registry through a Prometheus endpoint
In order to reduce the number of loaded agents and simplify the stack, we'd need to expose the Cassandra/DSE metrics registry in the mgmt api agent, using the following prototype as a base: https://github.com/burmanm/mcac-revisit

The new endpoint should be disabled by default and enabled through an env variable: `MGMT_API_METRICS_ENDPOINT=enabled`

The port on which metrics are exposed should also be configurable through an env variable, and default to port 9104 if the env variable doesn't exist: `MGMT_API_METRICS_PORT=9104`

No metric filtering should be done in this first implementation.

### LINKED ISSUE #229: Filter metrics using the METRICS_FILTERS env variable
The new metrics endpoint added in #228, should filter metrics based on the METRIC_FILTERS env variable that k8ssandra-operator generates:

```
METRIC_FILTERS:deny:org.apache.cassandra.metrics.Table deny:org.apache.cassandra.metrics.table allow:org.apache.cassandra.metrics.table.live_ss_table_count allow:org.apache.cassandra.metrics.Table.LiveSSTableCount allow:org.apache.cassandra.metrics.table.live_disk_space_used allow:org.apache.cassandra.metrics.table.LiveDiskSpaceUsed allow:org.apache.cassandra.metrics.Table.Pending allow:org.apache.cassandra.metrics.Table.Memtable allow:org.apache.cassandra.metrics.Table.Compaction allow:org.apache.cassandra.metrics.table.read allow:org.apache.cassandra.metrics.table.write allow:org.apache.cassandra.metrics.table.range allow:org.apache.cassandra.metrics.table.coordinator allow:org.apache.cassandra.metrics.table.dropped_mutations
```

The rules should apply as they did in MCAC and the filter names should be converted (basically replacing `.` with `_`) to match the names of the metrics generated by this new endpoint.

### LINKED ISSUE #233: Disable MCAC in our images
After merging the MCAC replacement, we need to make sure the MCAC itself is disabled if present in the images we build. In the OSS builds, we can simply remove its java-agent, but in the DSE builds we must add a parameter to the startup envs:

``-Dinsights.default_mode=disabled``

https://docs.datastax.com/en/dse/6.8/dse-dev/datastax_enterprise/config/insightsMetricsProperties.html#insightsMetricsProperties__insights.default_mode

burmanm:
This replaces the MCAC in our images with a new metrics collector. At the moment, it answers on port 9000 and path /metrics (anything else will return 404).

* Missing SSL support, how do we wish to proceed with this one? mgmt-api has configurable SSL properties already.
* There's no gateway from mgmt-api-http-server -> agent-metrics. Do we want/need such for some purpose?
* There's no Netty chunking, buffered writing.

Hopefully I didn't forget anything else.. I tested with 4.0.6 again, but lets see if mgmt-api building process / tests break for some reason (since CDC agent will modify behavior).

Ouh, we probably want more tests.. this PR doesn't have them ported, at least not yet.

Fixes #233 
Fixes #229 
Fixes #228


management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricsTools.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public CassandraMetricsTools() {
+        metricDefinitions = new HashMap<>();
+    }
+
+    public static String getHostId() {
+        try {
+            return StorageService.instance.getLocalHostId();
+        } catch (NullPointerException npe) {
+            return "123456789";
```
adejanovski, line 53: Shouldn't we log something?
I assume this can happen before bootstrap completes, right?

management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricsTools.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    public static String getClusterName() {
+        try {
+            return DatabaseDescriptor.getClusterName();
+        } catch (NullPointerException npe) {
+            return "Test Cluster";
```
adejanovski, line 61: Not sure about the implications here either. Does this mean we can emit metrics using "Test Cluster" as name and then with the actual cluster name once the node has fully bootstrapped?

management-api-agent-common/src/main/java/io/k8ssandra/metrics/config/ConfigReader.java
```diff
@@ -0,0 +1,34 @@
...
+public class ConfigReader {
+    public static final String CONFIG_PATH_PROPERTY = "collector-config-path";
+    public static final String CONFIG_PATH_DEFAULT = "/opt/management-api/configs/metric-collector.yaml";
+
+    public static Configuration readConfig() {
+        // Check env variable if there's any changes to the config path
+        String configPath = System.getProperty(CONFIG_PATH_PROPERTY);
+        if(configPath == null) {
+            configPath = CONFIG_PATH_DEFAULT;
+        }
```
adejanovski, line 18: nit: could be refactored into a ternary

management-api-agent-common/src/main/java/io/k8ssandra/metrics/config/ConfigReader.java
```diff
@@ -0,0 +1,34 @@
...
+
+        ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+        File configFile = new File(configPath);
+
+        // FileNotFoundException should be thrown if override is used
+        if(configFile.exists() || !configPath.equals(CONFIG_PATH_DEFAULT)) {
```
adejanovski, line 24: This condition isn't easy to understand.
Why don't we read the config if it's under the default path? Is it empty by default? Can't we have the case where we have some custom config loaded under the default path?

management-api-agent-common/src/main/java/io/k8ssandra/metrics/http/NettyHttpInitializer.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @Override
+    public void initChannel(SocketChannel ch) {
+        ChannelPipeline p = ch.pipeline();
+        if (sslCtx != null) {
+            p.addLast(sslCtx.newHandler(ch.alloc()));
+        }
+        p.addLast(new HttpServerCodec());
+        // TODO Do we need chunking for larger /metrics ?
```
adejanovski, line 24: I think we should account for clusters with 100s of tables and implement whatever we can to make this work for these as well.
How hard is chunking to implement?
We could probably create a follow up ticket for this.

management-api-agent-common/src/main/java/io/k8ssandra/metrics/http/NettyMetricsHttpServer.java
```diff
@@ -0,0 +1,39 @@
...
+import io.netty.handler.ssl.util.SelfSignedCertificate;
+
+import javax.net.ssl.SSLException;
+import java.security.cert.CertificateException;
+
+public class NettyMetricsHttpServer {
+
+    public void start(EventLoopGroup group) {
+        // Configure SSL.
+//        final SslContext sslCtx = buildSslContext();
```
adejanovski, line 19: Vector is loaded as a sidecar, and I think we can restrict metrics exposition to localhost only.
Maybe that would secure things enough that we don't have to deal with this?
If we implement this, can we re-use the Cassandra encryption settings?

management-api-agent-common/src/main/java/io/k8ssandra/metrics/http/NettyServerHandler.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    protected void channelRead0(ChannelHandlerContext ctx, HttpObject httpObject) throws Exception {
+        if (httpObject instanceof HttpRequest) {
+            HttpRequest req = (HttpRequest) httpObject;
+
+            URI uri = new URI(req.uri());
+            if (!uri.getPath().equals("/metrics")) {
+                // Send 404?
```
adejanovski, line 33: yes 👍

scripts/docker-entrypoint.sh
```diff
@@ -86,27 +81,6 @@ if [ "$1" = 'mgmtapi' ]; then
...
-        if [[ -n "$METRIC_FILTERS"  ]]; then
-            filter_array=(`echo $METRIC_FILTERS | sed 's/ /\n/g'`)
-            echo "filtering_rules:" >> ${MCAC_PATH}/config/metric-collector.yaml
-            for filter in "${filter_array[@]}"; do
-                echo "  - policy: $(echo ${filter} | cut -d':' -f1)" >> ${MCAC_PATH}/config/metric-collector.yaml
-                echo "    pattern: $(echo ${filter} | cut -d':' -f2)" >> ${MCAC_PATH}/config/metric-collector.yaml
-                echo "    scope: global" >> ${MCAC_PATH}/config/metric-collector.yaml
-            done
-        fi
-    fi
```
adejanovski, line 108: I'd like to keep it for now and make it possible to disable it instead.
We could selectively enable old MCAC or new MCAC, and use the same port for both, which would allow switching from one to the other without changing the ServiceMonitors in k8ssandra-operator.

adejanovski (commented):
> Missing SSL support, how do we wish to proceed with this one? mgmt-api has configurable SSL properties already.

Not sure. Maybe we can skip this in the initial implementation and create a follow up ticket to reduce the scope.

> There's no gateway from mgmt-api-http-server -> agent-metrics. Do we want/need such for some purpose?

I don't think so.

> There's no Netty chunking, buffered writing.

I added a comment on this. We may want to estimate how big the metrics can be with 100s of tables and handle chunking in a follow up ticket.

> Ouh, we probably want more tests.. this PR doesn't have them ported, at least not yet.

Yeah, we'll definitely need those before merging 👍

management-api-agent-common/src/main/java/com/datastax/mgmtapi/Agent.java
```diff
@@ -44,6 +38,9 @@ public static void premain(String arg, Instrumentation inst) throws Exception {
                 .transform(CassandraRoleManagerInterceptor.transformer())
                 .type(SystemDistributedReplicationInterceptor.type())
                 .transform(SystemDistributedReplicationInterceptor.transformer())
+                // Metrics server
+//                .type(MetricsInterceptor.type())
+//                .transform(MetricsInterceptor.transformer())
```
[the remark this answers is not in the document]
burmanm: Because we can't use two different interceptors this way in JDK8.

management-api-agent-common/src/main/java/io/k8ssandra/metrics/http/NettyMetricsHttpServer.java
```diff
@@ -0,0 +1,39 @@
...
+
+        ServerBootstrap b = new ServerBootstrap();
+        ServerBootstrap channel = b.group(group)
+                .childHandler(new NettyHttpInitializer(null))
+//                .handler(new LoggingHandler(LogLevel.DEBUG))
```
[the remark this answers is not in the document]
burmanm: We get Netty's logs indicating when channel is processed, such as ``READ READY [0x..]``. They're only useful for debugging the handler's process, but would make debugging everything else more difficult. Just enable it if you're developing the handler, otherwise - no.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/interceptors/CassandraDaemonInterceptor.java
```diff
@@ -90,9 +87,34 @@ public static void intercept(@SuperCall Callable<Void> zuper) throws Exception {
...
+
+            // Create /metrics handler. Note, this doesn't support larger than nThreads=1
+            final EventLoopGroup httpGroup = new EpollEventLoopGroup(1);
+
+            // Share them from HTTP server
+            NettyMetricsHttpServer server = new NettyMetricsHttpServer();
+            server.start(httpGroup);
+
+            logger.info("Metrics collector started");
+
```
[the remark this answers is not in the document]
burmanm: Do we have a reason to disable it? Just don't call the /metrics endpoint?

management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricRegistryListener.java
```diff
@@ -0,0 +1,300 @@
...
+        prototype.getDefinitions().removeIf(next -> !metricFilter.matches(next, dropwizardName));
+
+        if (prototype.getDefinitions().size() < 1) {
+            return;
+        }
+
+        RefreshableMetricFamilySamples familySamples;
+        if (!familyCache.containsKey(metricName)) {
+            familyCache.put(metricName, prototype);
+            cache.put(dropwizardName, metricName);
```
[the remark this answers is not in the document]
burmanm: You could do it multiple times, sure. But I don't see how it would conflict. Why would same metric be registered twice in different threads? And if they're different, which one would be correct? If they're the same, then no problem.

management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricsTools.java
burmanm: Yes, this happens before bootstrap. It actually NPEs only in the tests.

management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricsTools.java
burmanm: No, this is for test.

management-api-agent-common/src/main/java/io/k8ssandra/metrics/config/ConfigReader.java
burmanm: Those are ugly and hard to read. I only use them when I want different bytecode to be emitted ;)

management-api-agent-common/src/main/java/io/k8ssandra/metrics/config/ConfigReader.java
burmanm: The config does not exists by default, so in that case we don't want to throw Exception even if it can't be read. If the path has been modified, I'd expect the file to exists.

management-api-agent-common/src/main/java/io/k8ssandra/metrics/http/NettyHttpInitializer.java
burmanm: Chunking isn't, but I'm not sure if Prometheus scraper understands it. I don't think the 100s of tables is an issue here, as long as the reading client can handle it. 

We anyway process it all in the same moment.

management-api-agent-common/src/main/java/io/k8ssandra/metrics/http/NettyMetricsHttpServer.java
burmanm: Vector isn't the only scraper, so listening only in localhost isn't a good solution.

We could secure this with kube-rbac-proxy or similar, we have examples / settings in our kustomize templates already.

scripts/docker-entrypoint.sh
burmanm: I ran into issues with MCAC messing up the interceptor code (I guess same as Miles ran into with CDC). Will have to revisit this with more testing.

management-api-agent-common/src/main/java/io/k8ssandra/metrics/http/NettyMetricsHttpServer.java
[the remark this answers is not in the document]
adejanovski: gotcha 👍

management-api-agent-common/src/main/java/com/datastax/mgmtapi/interceptors/CassandraDaemonInterceptor.java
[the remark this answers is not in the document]
adejanovski: I assumed starting this new http server could have a footprint in the jvm for example? Maybe it's small enough that we don't care.

Dockerfile-4_0
```diff
@@ -27,12 +27,6 @@ COPY management-api-common ./management-api-common
 COPY management-api-server ./management-api-server
 RUN mvn -q -ff package -DskipTests
 
-# Download and extract Metrics Collector
-ENV MCAC_PATH /opt/metrics-collector
-RUN mkdir ${MCAC_PATH} && \
-    if test ! -e datastax-mcac-agent-${METRICS_COLLECTOR_VERSION}.tar.gz; then curl -L -O "https://github.com/datastax/metric-collector-for-apache-cassandra/releases/download/v${METRICS_COLLECTOR_VERSION}/datastax-mcac-agent-${METRICS_COLLECTOR_VERSION}.tar.gz"; fi && \
-    tar --directory ${MCAC_PATH} --strip-components 1 --gzip --extract --file datastax-mcac-agent-${METRICS_COLLECTOR_VERSION}.tar.gz
-
```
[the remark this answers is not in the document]
burmanm: We should not keep the MCAC enabled for too long. As long as it is running, it will do the metrics fetching even if nothing ever polls it. 

So, while we might have a working /metrics endpoint, those with large amount of tables are going to suffer as long as the MCAC is enabled, even if they never poll MCAC anymore.

scripts/docker-entrypoint.sh
burmanm: And it seems it still continues. MCAC will simply fail the job.

burmanm: That's because MCAC uses the age-old-Bytebuddy to do the interceptor (or some other legacy method) and that messes up with the updated one in this PR.

adejanovski (approved):
I've successfully tested this and reviewed the code.
The only thing left is to get CI to pass.

adejanovski:
Feel free to merge this PR if you're done with the changes after rebasing to fix the conflict.

PR #241, opened 2023-01-09
Allow datacenter names to have spaces which are replaced by __ to fix issues due to Cassandra startup scripts

### LINKED ISSUE #242: Support DC names containing spaces for replication jvm opts
Working on https://github.com/k8ssandra/cass-operator/issues/362, it appeared that if DC names have spaces, there's no way to pass them in -Dcassandra.system_distributed_replication_dc_names without Cassandra's startup failing.
This is likely due to how Cassandra's startup script build the command line.

We need a workaround to allow such DC names to be supported by the `-Dcassandra.system_distributed_replication_dc_names` and `-Dcassandra.system_distributed_replication`  jvm options.

adejanovski:
Fixes #242 

Working on https://github.com/k8ssandra/cass-operator/issues/362, it appeared that if DC names have spaces, there's no way to pass them in `-Dcassandra.system_distributed_replication_dc_names` without Cassandra's startup failing.
This is likely due to how Cassandra's startup script build the command line.

This PR allows to put `__` (double underscode) characters which will get replaced by a single space in the management API, so that we can support such DC names while not breaking Cassandra's startup.


adejanovski:
Closing in favor of a more future proof solution that is yet to be designed.

PR #249, opened 2023-01-19
Add DSE UBI images to matrix

### LINKED ISSUE #248: Add UBI images for DSE
Currently, DSE images are only built from Ubuntu based images. We should also include UBI based images.

emerkle826:
Fixes #248

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -199,9 +199,6 @@ jobs:
     if: ${{ github.ref == 'refs/heads/master' && github.event_name == 'push'}}
     needs: build-dse
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        jdk-version: [ubi7]
     steps:
       - name: Check out source code
         uses: actions/checkout@v3
@@ -232,7 +229,7 @@ jobs:
         uses: docker/metadata-action@v3
         with:
           images: datastax/dse-mgmtapi-6_8
-          tags: type=sha,prefix=dse68-${{ matrix.jdk-version }}-
+          tags: type=sha,prefix=dse68-ubi7-
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
       - name: Login to DockerHub
@@ -247,7 +244,7 @@ jobs:
         id: docker_build
         uses: docker/build-push-action@v3
         with:
-          file: dse-68/Dockerfile.${{ matrix.jdk-version }}
+          file: dse-68/Dockerfile.ubi7
           build-args: VERSION=dse68-jdk8-${{ steps.vars.outputs.sha_short }}
           context: .
           push: true
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,7 +48,7 @@ jobs:
         run: |
           mvn -B -q install --file pom.xml -Drun311tests=${{ matrix.run311tests }} -Drun40tests=${{ matrix.run40tests }} -Drun41tests=${{ matrix.run41tests }}
   build-dse:
-    if: ${{ github.ref == 'refs/heads/master' && github.event_name == 'push'}}
+    if: github.repository_owner == 'k8ssandra'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -48,7 +48,7 @@ jobs:
         run: |
           mvn -B -q install --file pom.xml -Drun311tests=${{ matrix.run311tests }} -Drun40tests=${{ matrix.run40tests }} -Drun41tests=${{ matrix.run41tests }}
   build-dse:
-    if: ${{ github.ref == 'refs/heads/master' && github.event_name == 'push'}}
```
Miles-Garnsey, line 51: Would we maybe not want to combine the repo owner condition with the old conditions rather than replacing?

.github/workflows/ci.yaml
```diff
@@ -185,9 +185,70 @@ jobs:
...
+          file: dse-68/Dockerfile.${{ matrix.jdk-version }}
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          platforms: linux/amd64
+          target: dse68
+
+  publish-dse-ubi:
+    name: Publish DSE UBI image
+    if: ${{ github.ref == 'refs/heads/master' && github.event_name == 'push'}}
```
Miles-Garnsey, line 199: Do you want to make this push only if it is in our fork?

.github/workflows/ci.yaml
```diff
@@ -185,9 +185,70 @@ jobs:
...
+    needs: build-dse
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        jdk-version: [ubi7]
```
Miles-Garnsey, line 204: Wrong variable name I reckon. Also can we have JDK 11 and 9 versions of UBI7 or no?

.github/workflows/ci.yaml
```diff
@@ -185,9 +185,70 @@ jobs:
...
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          platforms: linux/amd64
+          target: dse68
+
+  publish-dse-ubi:
+    name: Publish DSE UBI image
+    if: ${{ github.ref == 'refs/heads/master' && github.event_name == 'push'}}
+    needs: build-dse
```
Miles-Garnsey, line 200: Shouldn't this be `needs: build-dse-68-ubi7`

Miles-Garnsey (approved):
I'm approving to get out of your way, I've left a few comments/concerns around though which I think you should have a look at.

I appreciate that the only way to test this is to merge it, so I'll leave it to your discretion as to whether you want to just pull the trigger.

.github/workflows/ci.yaml
emerkle826: Actually, the jobs run on `push` or `pull_request`, except this one would only run on `push`. So now it runs on either, so long as the owner is k8ssandra. Which makes it like the OSS jobs.

.github/workflows/ci.yaml
emerkle826: That is the effective behavior because of the `needs` on the next line that won't run if the repo owner isn't k8ssandra. Provided I fix the next line as you pointed out.

.github/workflows/ci.yaml
emerkle826: Yep!

.github/workflows/ci.yaml
emerkle826: Yeah, the variable name probable should be changed here.
However, the official DSE images only use JDK8 for the UBI images, so I'm gonna go with "no" we can't have JDK 11 or 9 versions of UBI 7.

.github/workflows/ci.yaml
emerkle826: Actually, no this is correct, or at least the intent. `build-dse` is a job that builds 1 image (Ubuntu based JDK8 image) and runs the integration tests. The assumption here (and it is potentially a poor assumption) is that as long as 1 of the DSE images passes the DSE tests, they all should. So there is only 1 test job (`build-dse`) and if that passes, the image building jobs for all DSE variants are green lit to do the image building as those jobs don't run tests.

PR #250, opened 2023-01-20
Modify field name from source_labels to sourceLabels

Miles-Garnsey:
As discussed [here](https://github.com/k8ssandra/k8ssandra-operator/pull/813#issuecomment-1397939930).


management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/filter/FilteringSpec.java
```diff
@@ -10,24 +10,27 @@
 /**
  * Supports a subset of ServiceMonitorSpec's relabeling rules.
  *
- *             - source_labels: [subsystem, server]
- *               separator: "@"
- *               regex: "kata@webserver"
- *               action: "drop"
+ * - sourceLabels: [subsystem, server]
```
burmanm, line 13: Here is an example where your indent configs do not work. This automated indent messed up the comment's readability.

burmanm (approved):

PR #253, opened 2023-01-23
Add _sum to the timer histogram metric type

burmanm:


burmanm:
I'll push to upstream repo to get DSE build ..

PR #254, opened 2023-01-23
Add sum to timer, add logging to debug missing histogram

burmanm:


management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricRegistryListener.java
```diff
@@ -269,12 +269,14 @@ private void setTimerFiller(Timer timer, CassandraMetricDefinition proto, Cassan
 
                     buckets = (long[]) decayingHistogramOffsetMethod.invoke(snapshot);
                 } catch (InvocationTargetException | IllegalAccessException | NoSuchMethodException e) {
-                    throw new RuntimeException(e);
+                    logger.error(String.format("Unable to getOffsets for DSE, snapshotClass: %s", snapshotClass), e);
+//                    throw new RuntimeException(e);
```
adejanovski, line 273: Should this be thrown?
If not, could you remove that line?

adejanovski (approved):
Just one comment to remove (or uncomment), otherwise looks good 👍

management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricRegistryListener.java
burmanm: I'm not sure if we gain anything by throwing it. Even logging should perhaps be delegated to debug level.

adejanovski: If users cannot act on the problem and this is not a blocking error, then it should move at debug level indeed 👍 
You can remove the commented line if there's no need to re-throw the error.

PR #261, opened 2023-02-06
Remove the use of set-output in GHA and upgrade actions using deprecated deps

### LINKED ISSUE #247: `save-state` command is deprecated
The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

adejanovski:
Fixes #247

The create-release and upload-artifacts actions aren't maintained anymore and I've switched to the recommended maintained actions. This simplifies the workflow as we can create the release and upload artifacts in a single step.
I've checked that the @actions/core dependency was at least 1.10.0 as indicated in [GH's blog post](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

Successfully tested this by tweaking the jar release workflow trigger to be `test*.*.*` instead of `v*.*.*`:
Capture d’écran 2023-02-06 à 10 30 36

The release was created and the `jars.zip` artifact was successfully attached to it.
Here's the run output: https://github.com/k8ssandra/management-api-for-apache-cassandra/actions/runs/4102194244/jobs/7074849518

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -246,7 +246,7 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
       - name: Login to DockerHub
-        uses: docker/login-action@v1
+        uses: docker/login-action@v2
         with:
           username: ${{ secrets.DOCKER_HUB_USERNAME }}
           password: ${{ secrets.DOCKER_HUB_PASSWORD }}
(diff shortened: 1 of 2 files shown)


emerkle826 (approved):
This looks good. The only thing I can't remember now is if the [Docker release job](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/.github/workflows/docker-release.yaml) still uses GHAs that need to be updated as well

adejanovski:
I checked all the workflows, this is the only one that needed an update.

PR #263, opened 2023-02-07
Initial add of Cassandra trunk images

emerkle826:
This adds a UBI7 based Dockerfile that builds Cassandra from trunk and adds Management API to the image.

It currently has 1 limitation:
MCAC v0.3.4 is not compatible. (This can be disabled by launching the image with `MGMT_API_DISABLE_MCAC=true` set)


cassandra-oss/Dockerfile.ubi7
```diff
@@ -0,0 +1,183 @@
+# Copyright DataStax, Inc, 2020
+#   Please review the included LICENSE file for more information.
+#
+ARG BASETAG=7.8
+FROM registry.access.redhat.com/ubi7/ubi-minimal:${BASETAG} AS builder
+
+ARG URL_PREFIX=https://dlcdn.apache.org/cassandra
+ARG CASSANDRA_VERSION=trunk
```
emerkle826, line 8: This isn't used, but probably should be on line 115 below. It's set to `trunk` here because that's what it's going to build. I probably should adjust the git clone steps to checkout the branch that matches this version, that way it could be used to build any branch from the repo

cassandra-oss/Dockerfile.ubi7
```diff
@@ -0,0 +1,183 @@
...
+
+# COPY in files if they are available locally
+COPY * /
+WORKDIR /
+
+# Download and extract Metrics Collector
+ENV MCAC_PATH /opt/metrics-collector
+RUN mkdir ${MCAC_PATH} && \
+    if test ! -e datastax-mcac-agent-0.1.9.tar.gz; then curl -L -O "https://github.com/datastax/metric-collector-for-apache-cassandra/releases/download/v${METRICS_COLLECTOR_VERSION}/datastax-mcac-agent-${METRICS_COLLECTOR_VERSION}.tar.gz"; fi && \
+    tar --directory ${MCAC_PATH} --strip-components 1 --gzip --extract --file datastax-mcac-agent-${METRICS_COLLECTOR_VERSION}.tar.gz
```
emerkle826, line 28: As it turns out, the latest MCAC (v0.3.4) doesn't work with Cassandra trunk. So either I can pull this out, or we have to update MCAC to work with trunk.

cassandra-oss/Dockerfile.ubi7
```diff
@@ -0,0 +1,183 @@
...
+RUN set -x \
+    && rm -fr /etc/apt/sources.list.d/* \
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends git ant ant-optional make maven \
+    && git clone https://github.com/apache/cassandra.git \
+    && cd cassandra \
+    && ant artifacts mvn-install \
+    && mkdir ${CASSANDRA_PATH} \
+    && tar --directory ${CASSANDRA_PATH} --strip-components 1 --gzip --extract --file /build/cassandra/build/apache-cassandra-4.2-SNAPSHOT-bin.tar.gz \
```
emerkle826, line 50: the tar.gz bundle name is hard-coded, because that's what it happens to be today. Probably need to update this to dynamically figure it out based on what is built, possibly with a smart `find` command

cassandra-oss/Dockerfile.ubi7
```diff
@@ -0,0 +1,183 @@
...
+  mv ${CDC_AGENT_EDITION}-${CDC_AGENT_VERSION}-all.jar ${CDC_AGENT_PATH}/cdc-agent.jar
+
+############################################################
+
+FROM registry.access.redhat.com/ubi7/ubi-minimal:${BASETAG} as cass-trunk
+
+LABEL maintainer="DataStax, Inc <info@datastax.com>"
+LABEL name="Apache Cassandra"
+LABEL vendor="DataStax, Inc"
+LABEL release="${VERSION}"
```
emerkle826, line 115: `VERSION` isn't set anywhere, so this label will be empty. Either set it to `CASSANDRA_VERSION` from line 8 above, or dynamically figure out the version from the build process above

cassandra-oss/Dockerfile.ubi7
```diff
@@ -0,0 +1,183 @@
...
+    touch config/foo
+
+# Set user to run as
+USER cassandra:cassandra
+
+# Expose CASSANDRA folders
+VOLUME ["/config", "/var/lib/cassandra", "/var/log/cassandra"]
+
+# MANAGEMENT API, METRICS
+EXPOSE 8080 9103
```
emerkle826, line 176: @burmanm I should add the new non-mcac metrics port here as well, right?

cassandra-oss/files/entrypoint.sh
```diff
@@ -0,0 +1,201 @@
...
+        if [[ -n "$METRIC_FILTERS"  ]]; then
+            filter_array=(`echo $METRIC_FILTERS | sed 's/ /\n/g'`)
+            echo "filtering_rules:" >> ${MCAC_PATH}/config/metric-collector.yaml
+            for filter in "${filter_array[@]}"; do
+                echo "  - policy: $(echo ${filter} | cut -d':' -f1)" >> ${MCAC_PATH}/config/metric-collector.yaml
+                echo "    pattern: $(echo ${filter} | cut -d':' -f2)" >> ${MCAC_PATH}/config/metric-collector.yaml
+                echo "    scope: global" >> ${MCAC_PATH}/config/metric-collector.yaml
+            done
+        fi
+    fi
```
emerkle826, line 82: I guess this needs to be removed or MCAC needs to be updated to work on Cassandra trunk

PR #273, opened 2023-02-27
Replace tag value instead of adding a duplicate

burmanm:
If the tag name is already used in the relabeling process, replace th…e tag value - do not add a new one.

@Miles-Garnsey

### DIFF
--- a/management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricNameParser.java
+++ b/management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricNameParser.java
@@ -141,16 +141,17 @@ public class CassandraMetricNameParser {
           if (relabel.getTargetLabel().equals(RelabelSpec.METRIC_NAME_LABELNAME)) {
             metricDefinition.setMetricName(output);
           } else {
-            // Not the most effective way of doing this (map would be better - but perf is not an issue since this isn't done in the hotpath)
+            // Not the most effective way of doing this (map would be better - but perf is not an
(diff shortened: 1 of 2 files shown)


Miles-Garnsey:
Nice, thanks @burmanm. I was wondering if the labels were being duplicated. I think you have test failures, but I'll watch this PR and happy to review once the tests pass.

burmanm:
The 3.11 test failure is a known flake (there's a ticket for it), not related to this PR.

management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricNameParser.java
```diff
@@ -141,8 +141,20 @@ public void replace(String dropwizardName, CassandraMetricDefinition metricDefin
           if (relabel.getTargetLabel().equals(RelabelSpec.METRIC_NAME_LABELNAME)) {
             metricDefinition.setMetricName(output);
           } else {
-            metricDefinition.getLabelNames().add(relabel.getTargetLabel());
-            metricDefinition.getLabelValues().add(output);
+            // Not the most effective way of doing this (map would be better - but perf is not an
+            // issue since this isn't done in the hotpath)
+            boolean replaced = false;
```
Miles-Garnsey, line 146: I had a play with this. A map might have made life easier, but it isn't a big deal. 

Maybe consider doing something with `.stream()` for readability. There are some methods like `match` and `filter` that may do the job for you. It isn't a big deal either way IMO.

Miles-Garnsey (approved):

PR #278, opened 2023-03-02
Modify name parsing to do cleaning after every relabel rule

burmanm:
If the relabel rule modifies the _name__ field, clean it at that point also, instead of only at the end. Also, fix one bug in the default-metrics-settings for JVM relabel rule and add some additional tests

@Miles-Garnsey

### DIFF
--- a/management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricNameParser.java
+++ b/management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricNameParser.java
@@ -82,8 +82,8 @@ public class CassandraMetricNameParser {
 
     // Reclean with suffix added
     metricDef.setMetricName(
-            removeDoubleUnderscore(
-                    Collector.sanitizeMetricName(this.clean(metricDef.getMetricName()) + suffix)));
+        removeDoubleUnderscore(
+            Collector.sanitizeMetricName(this.clean(metricDef.getMetricName()) + suffix)));
 
     return metricDef;
   }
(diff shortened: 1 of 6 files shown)


Miles-Garnsey:
From earlier discussion with @burmanm:

There are a few different ways to tackle this problem. Right now it seems that at each stage of the metrics filtering pipeline in the cassandra metrics agent, there could be a change to the `__name__` field for some metrics where they end up with capital letters in the name.

Michael's solution here is to run a sanitisation after each rule is run (they happen serially). Other ways to tackle this (which might coexist with that solution) are:

1. Run the sanitisation once (only at the beginning of the default rules) so that all upper case characters are converted to lower case. (Not preferred, because some current rules rely on the capitals being in the metric names).
2. Run the sanitisation at the end of the default rules. (Preferred because the current rules don't need to change).

In either event, user defined rules occur after the default rules, so they would not be lower cased in a manner unexpected to the user.

@emerkle826 and @olim7t,  do either of you have any views on this topic? I think our current approach is to add a default sanitation rule at the end of the default rule chain. We will then either also sanitise in between each user rule, or only at the very end of processing user rules (because @burmanm is keen to ensure that user rules cannot break a prom scrape).

burmanm:
2. isn't possible in the sense of regexp, we can't write a rule for this since Java regexp doesn't support ``\L`` or ``\U``, these are Perl regexp extensions.

Miles-Garnsey:
> 2. isn't possible in the sense of regexp, we can't write a rule for this since Java regexp doesn't support `\L` or `\U`, these are Perl regexp extensions.

So there's no way to lower case the regexp outputs? In that case, let's just merge this PR!

Miles-Garnsey (approved):
Based on discussions, approving since we don't have other options.

PR #294, opened 2023-04-26
Parse tokens in endpoint states response

olim7t:


emerkle826 (approved):
This looks good. Thanks for fixing the tokens representation in the response!

emerkle826:
@olim7t If you are comfortable with the openapi doc commit I added, feel free to squash merge this.

PR #303, opened 2023-06-16
Allow cassandra-trunk docker image to be built from any repo/branch

yukim:
Added two build args:

 * `REPO`
 * `BRANCH`
 
to allow `cassandra-trunk` docker image to be built from any repo/branch.
For example, I used this change to build image from `cep-15-accord` branch to test Accord on kubernetes using cass-operator:

```
docker build -t ghcr.io/yukim/cass-management-api:cep-15-accord --build-arg="BRANCH=cep-15-accord" --file cassandra-trunk/Dockerfile.ubi8 --target cass-trunk --platform linux/amd64 .
```

### DIFF
--- a/cassandra-trunk/Dockerfile.ubi8
+++ b/cassandra-trunk/Dockerfile.ubi8
@@ -34,6 +34,8 @@ RUN if test ! -e tini-amd64; then curl -L -O "https://github.com/krallin/tini/re
 
 # Build and extract Cassandra
 FROM --platform=$BUILDPLATFORM azul/zulu-openjdk-debian:11 as cass-builder
+ARG REPO=https://github.com/apache/cassandra.git
+ARG BRANCH=trunk
 ARG COMMITSHA="HEAD"
 ENV CASSANDRA_PATH /opt/cassandra
 WORKDIR /build
@@ -42,7 +44,7 @@ RUN set -x \
     && rm -rf /var/lib/apt/lists/* \
     && apt-get update \
     && apt-get install -y --no-install-recommends git ant ant-optional make maven \
-    && git clone -b trunk --single-branch https://github.com/apache/cassandra.git \
+    && git clone -b ${BRANCH} --single-branch ${REPO} \
     && cd cassandra \
     && git checkout $(git rev-parse --short ${COMMITSHA}) \
     #&& ant -q -S echo-base-version > /build/cassandra.version \


cassandra-trunk/Dockerfile.ubi8
```diff
@@ -42,7 +44,7 @@ RUN set -x \
     && rm -rf /var/lib/apt/lists/* \
     && apt-get update \
     && apt-get install -y --no-install-recommends git ant ant-optional make maven \
-    && git clone -b trunk --single-branch https://github.com/apache/cassandra.git \
+    && git clone -b ${BRANCH} --single-branch ${REPO} \
```
emerkle826, line 47: So this will work as long as the branch you pick is based on a 5.0 version of Cassandra, and not a Release branch (i.e. the build artifact will be `5.0-SNAPSHOT`. If not, then you will need to handle:

1. https://github.com/yukim/management-api-for-apache-cassandra/blob/feature/cassandra-repo-arg/cassandra-trunk/Dockerfile.ubi8#L53

As the build artifact is assumed to be `5.0-SNAPSHOT`

2. https://github.com/yukim/management-api-for-apache-cassandra/blob/feature/cassandra-repo-arg/cassandra-trunk/Dockerfile.ubi8#L90-L97

You also have to adjust which artifacts of Management API are removed, kept and sym-linked in the section above. There are different artifacts built for different versions of Cassandra and things won't work if you include the wrong one.

3. https://github.com/yukim/management-api-for-apache-cassandra/blob/feature/cassandra-repo-arg/cassandra-trunk/Dockerfile.ubi8#L103

The `CDC_AGENT_EDITION` variable also needs to changed, depending on which version of Cassandra.

Having said all this, your changes should work as long as the branch is a 5.0-SNAPSHOT version. Seeing as how this Dockerfile is in the `cassandra-trunk` folder, I think that is a safe assumption for now.

I can approve this PR as-is, with all of the above caveats noted. However, it would be nice if you could figure out a way to deduce the Cassandra version so it's not hardcoded to `5.0-SNAPSHOT` here: https://github.com/yukim/management-api-for-apache-cassandra/blob/feature/cassandra-repo-arg/cassandra-trunk/Dockerfile.ubi8#L53

Let me know if you are willing to try this and I can wait to approve this PR then. If not, I think I can approve it now and this can be addressed later.

emerkle826 (changes requested):

PR #313, opened 2023-07-01
Add UBI based images for OSS Cassandra versions

### LINKED ISSUE #312: Add UBI based images for OSS Cassandra versions

emerkle826:
Fixes #312


Dockerfile-oss.ubi8
```diff
@@ -0,0 +1,163 @@
+ARG BASETAG=8.8
```
burmanm, line 1: question: Is there a reason we target certain minor-version 8.8 instead of using whatever is the newest RHEL-minor (for example when 8.9 comes out) ?

emerkle826: No reason other than to be a little more deliberate when we update to that new base release version. It is possible, though unlikely, that a minor version comes out with newer dependencies that cause issues. We can try to just refer to the latest major base version and see how it goes.

emerkle826: I've updated all the UBI based Dockerfiles so that the default is to pull the latest from the 8 series. When UBI 9 comes out, we will have to revisit this, but the latest 8.X will get picked up at build time unless `UBI_BASETAG` is set to a specific tag (i.e. 8.8 or something not `latest`)

PR #315, opened 2023-07-10
Add ARM64 image builds for DSE

emerkle826:


emerkle826:
@adejanovski If you have had a chance to test the image here:
https://hub.docker.com/repository/docker/erikmerkle/dse-mgmtapi-6_8/tags?page=1&ordering=last_updated
let me know if it works or not. If it does, we should be able to merge this.

adejanovski (approved):
Works great 👍

PR #317, opened 2023-07-11
Fix multi-arch DSE image builds

### LINKED ISSUE #318: Improve CI to better verify image builds
When a PR is opened, CI runs compile and test, but does not run the step to build and publish images to Docker. The problem is that the test phase only builds Linux ADM64 images, so ARM64 image builds aren't run/tested. We could improve the CI jobs so that multi-arch images are built every time CI runs, but only publish/push the images to Docker when the PR is merged to the master branch.

emerkle826:
Fixes #318

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -87,24 +87,27 @@ jobs:
         with:
           file: dse-68/Dockerfile.${{ matrix.jdk-version }}
           context: .
-          load: true
-          tags: ${{ steps.meta.outputs.tags }}
-          platforms: linux/amd64,linux/arm64
-          target: dse68
-      # Build UBI image in the same job after the Ubuntu image, only for JDK8
-      - name: Set outputs for DSE UBI image
-        if: ${{ matrix.jdk-version == 'jdk8' }}
-        id: vars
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-build.yaml
```diff
@@ -0,0 +1,178 @@
...
+#          build-args: VERSION=dse68-jdk8-${{ steps.vars.outputs.sha_short }}
+#          context: .
+#          push: false
+#          tags: ${{ steps.meta.outputs.tags }}
+#          platforms: linux/amd64,linux/arm64
+#          target: dse68
+#          cache-from: |
+#            type=local,src=${{ env.CACHE_PATH }}
+#            ${{ steps.dsemeta.outputs.tags }}
+#          cache-to: type=local,dest=${{ env.CACHE_PATH }}
```
emerkle826, line 178: This entire section(for DSE UBI images) is commented out because I haven't been able to find a way to build the DSE Ubuntu images in the previous job and make the images available here. The UBI images pull from the Ubuntu based images, and I can't figure out how to get the Docker Buildx cache transferred from the previous job into this job so that the UBI builds don't try to pull from Docker.

emerkle826: @adejanovski If this limitation is ok for now, I think this is good to go.

adejanovski (approved):

PR #321, opened 2023-07-19
Add OpenAPI Java client generation

### LINKED ISSUE #323: Generate Java client for OpenAPI spec
Generating a Java client based on the Management API spec will allow other projects to interact with Management API much easier (as opposed to downstream projects having to maintain their own REST client implementations that work with this project's API).

```[tasklist]
### DoD
- [ ] Add build task to generate Java client code that implements the[OpenAPI spec
- [ ] Client code should be built and published to a Maven artifactory repository during Management API release jobs
- [ ] Annotations on existing API endpoint implementations should be updated, as necessary
```

As a clarification for the last DoD item above, some of the endpoints are not currently annotated with a response `schema`. For example, [listKeyspaces](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-server/src/main/java/com/datastax/mgmtapi/resources/KeyspaceOpsResources.java#L234-L248) annotates the Response as JSON, but does not specify a `schema` attribute. This causes the client generation to create a method that has a `Void` return type, which will clearly not work for code using the client to retrieve a list of keyspaces. For these methods, adding the `schema` attribute will not change the implementation, so it won't affect existing Management API consumers. It only changes the OpenAPI spec that is generated, which will alter the client code generated.

emerkle826:
Fixes #323

### DIFF
--- a/.github/workflows/openapi-publish.yaml
+++ b/.github/workflows/openapi-publish.yaml
@@ -1,0 +1,44 @@
+name: Publish OpenAPI Client
+on:
+  workflow_dispatch:
+jobs:
+  publish-openapi:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v3
+      - name: Set up JDK 11
+        uses: actions/setup-java@v3
+        with:
+          java-version: 11
+          distribution: 'zulu'
+      - name: Cache Maven packages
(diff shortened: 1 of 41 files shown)


emerkle826:
Closing this in favor of splitting out the endpoint work into multiple PRs. See #324 and #325

PR #324, opened 2023-07-26
Add OpenAPI client generation

### LINKED ISSUE #323: Generate Java client for OpenAPI spec

emerkle826:
This PR does 2 things:
1. Renames the Maven artifact groupId from `com.datastax` to `io.k8ssandra`
2. Creates a workflow_dispatch GHA to on-demand build the OpenAPI Java client and publish it to Datasatx Artifactory

Fixes #323


Miles-Garnsey (approved):
I've taken a quick glance, it all looks good to me in theory! 

Issue: I couldn't see a release when I searched in artifactory, have you confirmed that this works by running the mvn deploy command locally? I don't have the credentials, so my ability to test this is a bit limited.

Suggestion: Also, when I look at the other PR you have open, it skips the openAPI build in the docker files. Should those changes be a part of this PR instead? The docs should probably also be in this one. 

Issue: Is there a changelog update that needs to happen here too?

emerkle826:
> Issue: I couldn't see a release when I searched in artifactory, have you confirmed that this works by running the mvn deploy command locally? I don't have the credentials, so my ability to test this is a bit limited.

I haven't tried it manually, so I will.

emerkle826:
> Suggestion: Also, when I look at the other PR you have open, it skips the openAPI build in the docker files. Should those changes be a part of this PR instead? The docs should probably also be in this one.

The Dockerfile changes probably should be in here. It's not critical as even if the Dockerfiles generate the client code, it's not included in the final image. I disabled the generation in there to trim out something else that doesn't need to happen, but it isn't much of a time or space waster in the grand scheme of things.

I will move that in here though and I will move/add the documentation updates to this one as well.

emerkle826:
> Issue: Is there a changelog update that needs to happen here too?

Yep, that's why we have reviews :smile:

emerkle826:
> Issue: I couldn't see a release when I searched in artifactory, have you confirmed that this works by running the mvn deploy command locally? I don't have the credentials, so my ability to test this is a bit limited.

@Miles-Garnsey I have successfully published a version of the client based on this PR and it works. I have updated the README [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/324/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R430-R477) to show how to add it to a project.

emerkle826:
@Miles-Garnsey @adejanovski I'm going to go ahead and merge this. The core of what the PR needs to do is publish the client. It can do that with an on-demand workflow, and it will do that on release. When we complete discussions on if/where to publish other artifacts, we can easily PR that work in as well.

PR #333, opened 2023-07-28
Fix OpenAPI annotations for existing endpoints

### LINKED ISSUE #332: Fix OpenAPI/Swagger annotations
Some of the API endpoint methods are not correctly annotated with respect to responses. When the OpenAPI document is generated from these annotations, it generates a document that does not properly or fully describe the endpoints. This document is used to generate a Java client that can be used to interact with the API, however this deficiency causes the client to be generated with missing/incorrect response types/objects.

For example, the "listKeyspaces" endpoint is currently documented here:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-server/doc/openapi.json#L411-L433
```json
    "/api/v0/ops/keyspace" : {
      "get" : {
        "operationId" : "listKeyspaces",
        "parameters" : [ {
          "in" : "query",
          "name" : "keyspaceName",
          "schema" : {
            "type" : "string"
          }
        } ],
        "responses" : {
          "200" : {
            "content" : {
              "application/json" : {
                "example" : [ "system_schema", "system", "system_auth", "system_distributed", "system_traces" ]
              }
            },
            "description" : "List of Keyspaces"
          }
        },
        "summary" : "List the keyspaces existing in the cluster"
      }
    }
```

While the `example` does show that the response should contain an array of Strings, the schema for the response is missing. This causes the Java client generator to produce a method that has a `Void` return type. This makes the client impossible to use to retrieve the list of keyspaces.

emerkle826:
Fixes #332

### DIFF
--- a/management-api-server/doc/openapi.json
+++ b/management-api-server/doc/openapi.json
@@ -355,7 +355,10 @@
           "200" : {
             "content" : {
               "text/plain" : {
-                "example" : "OK"
+                "example" : "OK",
+                "schema" : {
+                  "type" : "string"
+                }
               }
             },
             "description" : "Role created"
@@ -363,7 +366,10 @@
           "400" : {
             "content" : {
               "text/plain" : {
-                "example" : "Username is empty"
+                "example" : "Username is empty",
+                "schema" : {
+                  "type" : "string"
+                }
               }
             },
             "description" : "Username and/or password is empty"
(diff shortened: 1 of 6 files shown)


burmanm (approved):
Not too many typed types here, so string it is..

emerkle826:
> Not too many typed types here, so string it is..

Yes, the goal with this PR is to just fix the annotations so that they correctly document the existing response behavior. That is step 1. Step 2 is to add new endpoints that we need for Reaper where these string responses can be refactored into proper Java objects. There are a lot of not-really-designed REST endpoints in here.

PR #334, opened 2023-07-28
Skip OpenAPI generation in Dockerfiles

emerkle826:
Some of these were left out in the initial OpenAPI client publishing PR


Miles-Garnsey (approved):
My understanding of this PR is that `DskipOpenApi` just skips the creation of the OpenAPI spec, which is not required to build the Cassandra/DSE images (which include the management API) in any event. 

On that basis this is approved, please let me know if I'm misinterpreting the purpose of this PR.

emerkle826:
> On that basis this is approved, please let me know if I'm misinterpreting the purpose of this PR.

That is correct. Trying to keep the scope/concern of each of these PRs limited so that it's not so much to review in one PR.

PR #341, opened 2023-08-10
Add artifact deployment to workflow_dispatch to keep client and agents in sync

### LINKED ISSUE #339: OpenAPI client publish does not also publish other artifacts
The on-demand OpenAPI client publish job publishes the latest client jarfile to a Maven repository. Latest in this context is the most recent commit of the branch from which the job is run. The problem is that this client won't necessarily be compatible with the latest released Agent and Server jarfiles of Management API. 

```[tasklist]
### Definition of Done
- [ ] Publish job also publishes versions of the agents and the server jarfiles
```

emerkle826:
Fixes #339 

This will publish all of the project Maven artifacts at the same time as the Client jar so that downstream projects can access the same version of the agent(s) and server files as the client jarfile.


Miles-Garnsey:
Suggestion: comment out artifactory related jobs and make a comment that they don't work at the moment.

Miles-Garnsey:
@emerkle826 looks like we have failing tests here, happy to review once they're passing!

emerkle826:
@Miles-Garnsey A re-run of the tests and they are all passing now. At some point I will dig into the flaky one as it's not always the same test, but usually the same symptom....

Miles-Garnsey:
We've validated that manually running the `Publish OpenAPI client` does push both client and agent jars to Cloudsmith. 

The workflow that is run is the one on the branch, not the one on main, so we have been able to validate this prior to merge.

Erik has confirmed that no rebase is applied to the branch prior to the new publication CI jobs running, so the commit sha from HEAD is the correct one to apply to the artifacts.

The only thing remaining prior to approval is to comment out the DataStax publication target, since that isn't working at present.

Miles-Garnsey:
DataStax publication option is removed. This is ready for merge pending green CI.

Miles-Garnsey (approved):
CI is green, approving and merging.

PR #346, opened 2023-08-16
Add getClusterName endpoint, add currentStreams DTO for reloadLocalSchema

### LINKED ISSUE #345: Implement host related methods
## Definition of Done
- [ ] `PartitionerName` is available via an endpoint
- [ ] `ClusterName` is available via an endpoint
- [ ] `CurrentStreams` is available via an endpoint

emerkle826:
Fixes #345

### DIFF
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/resources/MetadataResources.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/resources/MetadataResources.java
@@ -102,20 +102,4 @@ public class MetadataResources extends BaseResources {
           return Response.ok(featureSet).build();
         });
   }
-
-  @GET
-  @Path("/clustername")
-  @Operation(summary = "Returns the Cassandra cluster name", operationId = "getClusterName")
-  @Produces(MediaType.TEXT_PLAIN)
-  @ApiResponse(
-      responseCode = "200",
-      description = "Cassandra Cluster Name",
-      content =
-          @Content(
-              mediaType = MediaType.TEXT_PLAIN,
-              schema = @Schema(implementation = String.class),
-              examples = @ExampleObject(value = "myCluster")))
-  public Response getClusterName() {
-    return executeWithStringResponse("CALL NodeOps.getClusterName()");
-  }
 }
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/resources/MetadataResources.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/resources/MetadataResources.java
@@ -35,7 +35,7 @@ public class MetadataResources extends BaseResources {
   @Produces(MediaType.TEXT_PLAIN)
   @ApiResponse(
       responseCode = "200",
-      description = "Cassandra version'",
+      description = "Cassandra version",
       content =
           @Content(
               mediaType = MediaType.TEXT_PLAIN,
@@ -101,5 +101,21 @@ public class MetadataResources extends BaseResources {
           FeatureSet featureSet = new FeatureSet(cassandraVersion, "");
           return Response.ok(featureSet).build();
         });
+  }
+
+  @GET
+  @Path("/clustername")
+  @Operation(summary = "Returns the Cassandra cluster name", operationId = "getClusterName")
+  @Produces(MediaType.TEXT_PLAIN)
+  @ApiResponse(
+      responseCode = "200",
+      description = "Cassandra Cluster Name",
+      content =
+          @Content(
+              mediaType = MediaType.TEXT_PLAIN,
+              schema = @Schema(implementation = String.class),
+              examples = @ExampleObject(value = "myCluster")))
+  public Response getClusterName() {
+    return executeWithStringResponse("CALL NodeOps.getClusterName()");
   }
 }
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
@@ -11,6 +11,7 @@ import com.datastax.mgmtapi.ManagementApplication;
 import com.datastax.mgmtapi.resources.common.BaseResources;
 import com.datastax.mgmtapi.resources.helpers.ResponseTools;
 import com.datastax.mgmtapi.resources.models.RepairRequest;
+import com.datastax.mgmtapi.resources.models.StreamingInfo;
 import com.datastax.mgmtapi.resources.models.TakeSnapshotRequest;
 import com.datastax.oss.driver.api.core.cql.Row;
 import com.google.common.collect.ImmutableList;
@@ -329,8 +330,7 @@ public class NodeOpsResources extends BaseResources {
       content =
           @Content(
               mediaType = MediaType.APPLICATION_JSON,
-              schema = @Schema(implementation = String.class),
-              examples = @ExampleObject(value = STREAMING_INFO_RESPONSE_EXAMPLE)))
+              schema = @Schema(implementation = StreamingInfo.class)))
   @Operation(summary = "Retrieve Streaming status information", operationId = "getStreamInfo")
   public Response getStreamInfo() {
     return handle(
@@ -608,33 +608,6 @@ public class NodeOpsResources extends BaseResources {
               .build();
         });
   }
-
-  private static final String STREAMING_INFO_RESPONSE_EXAMPLE =
-      "{\n"
-          + "    \"entity\": [],\n"
-          + "    \"variant\": {\n"
-          + "        \"language\": null,\n"
-          + "        \"mediaType\": {\n"
-          + "            \"type\": \"application\",\n"
-          + "            \"subtype\": \"json\",\n"
-          + "            \"parameters\": {},\n"
-          + "            \"wildcardType\": false,\n"
-          + "            \"wildcardSubtype\": false\n"
-          + "        },\n"
-          + "        \"encoding\": null,\n"
-          + "        \"languageString\": null\n"
-          + "    },\n"
-          + "    \"annotations\": [],\n"
-          + "    \"mediaType\": {\n"
-          + "        \"type\": \"application\",\n"
-          + "        \"subtype\": \"json\",\n"
-          + "        \"parameters\": {},\n"
-          + "        \"wildcardType\": false,\n"
-          + "        \"wildcardSubtype\": false\n"
-          + "    },\n"
-          + "    \"language\": null,\n"
-          + "    \"encoding\": null\n"
-          + "}";
 
   private static final String SNAPSHOT_DETAILS_RESPONSE_EXAMPLE =
       "{\n"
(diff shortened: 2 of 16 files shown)


management-api-server/src/main/java/com/datastax/mgmtapi/resources/MetadataResources.java
```diff
@@ -102,4 +102,20 @@ public Response getFeatureSet() {
...
+  @Path("/clustername")
+  @Operation(summary = "Returns the Cassandra cluster name", operationId = "getClusterName")
+  @Produces(MediaType.TEXT_PLAIN)
+  @ApiResponse(
+      responseCode = "200",
+      description = "Cassandra Cluster Name",
+      content =
+          @Content(
+              mediaType = MediaType.TEXT_PLAIN,
+              schema = @Schema(implementation = String.class),
```
Miles-Garnsey, line 116: Isssue: Can we have this wrapped in JSON please :)

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/BaseEntity.java
```diff
@@ -0,0 +1,213 @@
...
+ */
+package com.datastax.mgmtapi.resources.models;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+public abstract class BaseEntity {
```
Miles-Garnsey, line 14: That's what I was thinking of. Nice work.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/EndpointStates.java
```diff
@@ -11,26 +11,11 @@
...
-
-  @JsonProperty(value = "language", required = false)
-  public final String language;
-
-  @JsonProperty(value = "encoding", required = false)
-  public final String encoding;
-
   @JsonCreator
   public EndpointStates(
       @JsonProperty("entity") List<Map<String, String>> entity,
```
Miles-Garnsey, line 21: This isn't ideal. Are we sure we can't make this any more strongly typed right now with enums for the states? Will that break backwards compatibility somehow, or are we just trying to do this quickly (because we know that we're going to refactor all of it later anyway)?

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/BaseEntity.java
```diff
@@ -0,0 +1,213 @@
...
+      hash = 83 * hash + Objects.hashCode(this.type);
+      hash = 83 * hash + Objects.hashCode(this.subtype);
+      hash = 83 * hash + Objects.hashCode(this.parameters);
+      hash = 83 * hash + Objects.hashCode(this.wildcardType);
+      hash = 83 * hash + Objects.hashCode(this.wildcardSubtype);
+      return hash;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
Miles-Garnsey, line 126: Should we add a refactoring ticket to adopt Lombok into this project at some point?

management-api-server/src/test/java/com/datastax/mgmtapi/MetadataResourcesTest.java
```diff
@@ -50,6 +51,17 @@ public void testFeatureSetEndpoint() throws Exception {
...
   }
 
+  @Test
+  public void testGetClusterName() throws Exception {
+    K8OperatorResourcesTest.Context context = setup();
+    MockHttpResponse response = getMockHttpResponse(context, "/metadata/clustername");
+    assertEquals(HttpStatus.SC_OK, response.getStatus());
+    String clusterName = response.getContentAsString();
+    Assert.assertTrue(
+        "Clustername expected to containe \"myCluster\", but was " + clusterName,
```
Miles-Garnsey, line 61: Nit: typo `containe`

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/EndpointStates.java
Miles-Garnsey: I know we discussed this earlier this morning, just want to make sure I'm understanding our motivations for taking this path, given we can retrofit schemas into these endpoints right now.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/MetadataResources.java
```diff
@@ -102,4 +102,20 @@ public Response getFeatureSet() {
           return Response.ok(featureSet).build();
         });
   }
+
+  @GET
+  @Path("/clustername")
```
Miles-Garnsey, line 107: Suggestion: In many respects, I would rather have this returned from some endpoint like `/hosts` or `/endpoints` as a part of their response. However, given this is sitting in a v1 API, I'm not going to fuss about it.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
```diff
@@ -329,8 +330,7 @@ public Response reloadLocalSchema() {
       content =
           @Content(
               mediaType = MediaType.APPLICATION_JSON,
-              schema = @Schema(implementation = String.class),
-              examples = @ExampleObject(value = STREAMING_INFO_RESPONSE_EXAMPLE)))
+              schema = @Schema(implementation = StreamingInfo.class)))
```
Miles-Garnsey, line 333: Suggestion: Is `STREAMING_INFO_RESPONSE_EXAMPLE` still used anywhere? Can we remove that ugly block of text?

Miles-Garnsey (changes requested):
One or two issues, and a few questions. If I'm reading this right you're adding getClusterName and adding a `streamingInfo` model class which gets returned from `reloadLocalSchema`? 

Can we update the PR title to reflect that maybe?

management-api-server/src/main/java/com/datastax/mgmtapi/resources/MetadataResources.java
burmanm: I wouldn't mind adding to /endpoints as field as it doesn't break anything when new fields appear there (that happens naturally with Cassandra versions). As long as Cassandra itself doesn't add such field name.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/BaseEntity.java
burmanm: Please no, isn't Lombok already abandoned since most of it's benefits are in newer JDK?

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/EndpointStates.java
burmanm: This data is returned directly from Cassandra, so adding some enums is probably going to break at some point (unless we postprocess more from Cassandra before returning to users). We don't technically control what's returned by this endpoint, Cassandra does.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/BaseEntity.java
Miles-Garnsey: Maybe, but we aren't on a newer JDK...

emerkle826:
> One or two issues, and a few questions. If I'm reading this right you're adding getClusterName and adding a `streamingInfo` model class which gets returned from `reloadLocalSchema`?

Yes, adding getClusterName and a DTO for streaming info. However, the DTO is only returned from `getStreamInfo()`, not  `reloadLocalSchema`

management-api-server/src/main/java/com/datastax/mgmtapi/resources/MetadataResources.java
emerkle826: For this one, there isn't a need since a simple String will suffice. But actually this may be moot for now because I like the idea of sticking this information in the endpoint states as it would mean no new endpoint.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/BaseEntity.java
emerkle826: :+1:

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/EndpointStates.java
emerkle826: @burmanm is right. Also, doing anything with the enums specifically will break the existing contract of the endpoint, meaning it won't be backward compatible. In a future refactoring ticket, we can introduce more strong typing and handle the case where Cassandra changes, ads or deletes an enum, which should address all the concerns here. I haven't written those tickets yet, but I will.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/BaseEntity.java
emerkle826: I am not in favor of Lombok. If you want to add the ticket, I won't stop you :wink:

management-api-server/src/test/java/com/datastax/mgmtapi/MetadataResourcesTest.java
emerkle826: good catch

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
emerkle826: I think I did.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/BaseEntity.java
Miles-Garnsey: OK, no Lombok... Let's leave it :)

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/EndpointStates.java
Miles-Garnsey: I think there's a backwards compatible way to do this, but if you guys both agree to defer this work then let's keep moving forward.

Miles-Garnsey (approved):
All feedback addressed, still one failing UBI test but this can be resolved prior to merge to master. 

Approving merge to the integration branch now.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/BaseEntity.java
adejanovski: @Miles-Garnsey, don't forget to follow our [PR comments convention](https://docs.k8ssandra.io/contribute/pr_review/#comments-convention), thanks!

PR #348, opened 2023-08-16
Fix getSnapshotDetails endpoint

### LINKED ISSUE #347: Fix getSnapshot endpoint
The current endpoint returns a JSON String, but needs to be fitted with a matching DTO

```[tasklist]
### Definition of Done
- [ ] The OpenAPI/Swagger doc defines a DTO object returned by the getSnapshots endpoint
- [ ] The generated OpenAPI client returns the DTO object instead of raw JSON

### PR DESCRIPTION
**Pull request by emerkle826**
Fixes #347 

NOTE: This should wait for #345 to be merged as it contains a commit from that PR. Rebase this branch after that has been merged and this will reduce to a single commit adding the DTO.

### DIFF
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
@@ -11,6 +11,7 @@ import com.datastax.mgmtapi.ManagementApplication;
 import com.datastax.mgmtapi.resources.common.BaseResources;
 import com.datastax.mgmtapi.resources.helpers.ResponseTools;
 import com.datastax.mgmtapi.resources.models.RepairRequest;
+import com.datastax.mgmtapi.resources.models.SnapshotDetails;
 import com.datastax.mgmtapi.resources.models.StreamingInfo;
 import com.datastax.mgmtapi.resources.models.TakeSnapshotRequest;
 import com.datastax.oss.driver.api.core.cql.Row;
@@ -351,12 +352,12 @@ public class NodeOpsResources extends BaseResources {
   @Produces(MediaType.APPLICATION_JSON)
   @ApiResponse(
       responseCode = "200",
-      description = "Cassandra snapshot details",
+      description =
+          "Cassandra snapshot details. Use 'null' values for query parameters to exclude result filtering against the parameter.",
       content =
           @Content(
               mediaType = MediaType.APPLICATION_JSON,
-              schema = @Schema(implementation = String.class),
-              examples = @ExampleObject(value = SNAPSHOT_DETAILS_RESPONSE_EXAMPLE)))
+              schema = @Schema(implementation = SnapshotDetails.class)))
   @Operation(summary = "Retrieve snapshot details", operationId = "getSnapshotDetails")
   public Response getSnapshotDetails(
       @QueryParam("snapshotNames") List<String> snapshotNames,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by emerkle826** [comment]
@Miles-Garnsey same issue here with the failing tests. UBI image test will fail when new endpoints are added as the Ubuntu based images haven't been published yet with the endpoint.

**Comment by Miles-Garnsey** [comment]
Corresponding Reaper PR is [here](https://github.com/thelastpickle/cassandra-reaper/pull/1358). It is still WIP.

File: management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java [hunk 1]
```diff
@@ -355,8 +356,7 @@ public Response getStreamInfo() {
       content =
           @Content(
               mediaType = MediaType.APPLICATION_JSON,
-              schema = @Schema(implementation = String.class),
-              examples = @ExampleObject(value = SNAPSHOT_DETAILS_RESPONSE_EXAMPLE)))
+              schema = @Schema(implementation = SnapshotDetails.class)))
   @Operation(summary = "Retrieve snapshot details", operationId = "getSnapshotDetails")
   public Response getSnapshotDetails(
       @QueryParam("snapshotNames") List<String> snapshotNames,
```
Miles-Garnsey, line 362: We need a version of this method which takes no arguments. The current JMX implementation appears to simply return all snapshots. The parameters can be optional if that is easier/possible.

emerkle826: Looking into the OpenAPI code generator, it doesn't support any kind of Optional query parameter markings. The client code that is generated does take a list for both parameters (snapshot names and keyspaces), but calling the method with `null` will result in the client simply omitting the parameters. So in Reaper, you would do something like this:
```java
SnapshotDetails details = apiClient.getSnapshotDetails(null, null);
```
which would make the query to Management API and not provide either parameter. On the Management API side here, it interprets `null` as "get all of them".

Miles-Garnsey: Can we use the Optional<T> type? https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

Miles-Garnsey: At least then it is clear what the behaviour will be from the type system, if not in the docs via some annotation.

Miles-Garnsey: OK, I can see the issue with generating the 4 methods. I can also see that if we switch the types to Optional<List> for both parms then that's a breaking change, which we don't necessarily want to sign up to now. For v2 perhaps we can get better typing on the parameters so that methods 2 and 3 are distinguishable?

For now, could we get java doc on the method to explain this?

Miles-Garnsey: There is already an `optional` marking on the javadoc. Resolving.

Miles-Garnsey (approved):
Needs a rebase prior to being mergeable, approving based on all issues being resolved and the last test that ran being green. This is to be merged into the integration branch.

Miles-Garnsey:
Note: this has been manually tested.

PR #352, opened 2023-08-18
Add new listTables operation with more detailed response

olim7t:
Fixes: #361


management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -534,17 +534,18 @@ public Map<String, String> getReplication(@RpcParam(name = "keyspaceName") Strin
...
             .column("table_name")
+            .column("compaction")
             .where(Relation.column("keyspace_name").isEqualTo(QueryBuilder.literal(keyspaceName)))
             .asCql();
     UntypedResultSet rows = ShimLoader.instance.get().processQuery(query, ConsistencyLevel.ONE);
-    List<String> tables = new ArrayList<>();
+    List<Table> tables = new ArrayList<>();
     for (UntypedResultSet.Row row : rows) {
-      tables.add(row.getString("table_name"));
+      tables.add(new Table(row.getString("table_name"), row.getFrozenTextMap("compaction")));
```
olim7t, line 548: All that Reaper needs is the compaction class, but in case we need other fields later I think it's better to mimic the table's structure.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -534,17 +534,18 @@ public Map<String, String> getReplication(@RpcParam(name = "keyspaceName") Strin
     return rows.one().getMap("replication", UTF8Type.instance, UTF8Type.instance);
   }
 
-  @Rpc(name = "getTables")
-  public List<String> getTables(@RpcParam(name = "keyspaceName") String keyspaceName) {
+  @Rpc(name = "getTables", multiRow = true)
+  public List<Table> getTables(@RpcParam(name = "keyspaceName") String keyspaceName) {
```
olim7t, line 538: If I understood the code correctly, this will serialize each object as a row, and each public fields of the object as a column.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/TableOpsResources.java
```diff
@@ -326,7 +327,7 @@ public Response flush(KeyspaceRequest keyspaceRequest) {
       content =
           @Content(
               mediaType = MediaType.APPLICATION_JSON,
-              schema = @Schema(implementation = String.class),
+              array = @ArraySchema(schema = @Schema(implementation = String.class)),
```
olim7t, line 330: Not strictly related to this PR, but it was wrong.

olim7t (commented):
This is not complete yet, I still need to add the actual REST operation.

management-api-agent-3.x/src/main/java/com/datastax/mgmtapi/rpc/ObjectSerializer3x.java
```diff
@@ -73,7 +73,7 @@ public ObjectSerializer3x(Class<T> clazz, Type genericType) {
                             field -> field.getName(),
                             field ->
                                 new FieldSerializer(
-                                    GenericSerializer3x.getType(field.getType()), field))));
+                                    GenericSerializer3x.getType(field.getGenericType()), field))));
```
emerkle826, line 76: Issue:
We probably need this in the DSE 7 ObjectSerializer class as well. Unfortunately with the dependencies being internal Cassandra/DSE classes, we have to duplicate a lot of this to compile against the correct versions.

management-api-agent-3.x/src/test/java/com/datastax/mgmtapi/rpc/ObjectSerializer3xTest.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package com.datastax.mgmtapi.rpc;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+import java.util.Map;
+import org.junit.Test;
+
+public class ObjectSerializer3xTest {
```
emerkle826, line 14: Issue:
We should probably duplicate this in all the agent projects as well. There might be a way to only write a single test that is run for each version, but if it's faster, just duplicate the test in each project and maybe write a refactor ticket to consolidate in the future.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/Table.java
```diff
@@ -0,0 +1,26 @@
...
+  @JsonProperty(value = "compaction")
+  public final Map<String, String> compaction;
+
+  @JsonCreator
+  public Table(
+      @JsonProperty("name") String name,
+      @JsonProperty("compaction") Map<String, String> compaction) {
+    this.name = name;
+    this.compaction = compaction;
+  }
```
emerkle826, line 25: suggestion:
I know it's not necessary, but maybe implement `hashCode()`, `equals()` and `toString()`? I'm fine without those if you don't see any real value.

management-api-server/src/test/java/com/datastax/mgmtapi/BaseDockerIntegrationTest.java
```diff
@@ -32,6 +32,7 @@
 public abstract class BaseDockerIntegrationTest {
   protected static final Logger logger = LoggerFactory.getLogger(BaseDockerIntegrationTest.class);
   protected static final String BASE_PATH = "http://localhost:8080/api/v0";
+  protected static final String BASE_PATH_V2 = "http://localhost:8080/api/v2";
```
emerkle826, line 35: thought:
This seems like we don't have any of the `v1` endpoints covered in the tests. I'll create a ticket to add them later.

emerkle826 (changes requested):
Just a few minor changes.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/TableOpsResources.java
```diff
@@ -0,0 +1,81 @@
...
+      description = "Keyspace name not provided",
+      content =
+          @Content(
+              mediaType = MediaType.TEXT_PLAIN,
+              schema = @Schema(implementation = String.class),
+              examples =
+                  @ExampleObject(
+                      value = "List tables failed. Non-empty 'keyspaceName' must be provided")))
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Operation(summary = "List the table names in the given keyspace", operationId = "listTables")
```
emerkle826, line 59: Suggestion:
```suggestion
  @Operation(summary = "List the table names in the given keyspace", operationId = "listTablesV2")
```
So that the generated client creates a method called `listTablesV2` instead of leaving it to the client generation to avoid duplicate methods. Currently, the client generated will call this method `listTables1` which is not clear that it is actually a V2 implementation.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/TableOpsResources.java
```diff
@@ -0,0 +1,81 @@
...
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.http.HttpStatus;
+
+@Path("/api/v2/ops/tables")
+public class TableOpsResources extends BaseResources {
+
```
emerkle826, line 34: Issue:
I think we can move the `list()` method you have in here to the v1 `TableOpsResources` class here:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/feature/get-tables/management-api-server/src/main/java/com/datastax/mgmtapi/resources/v1/TableOpsResources.java
That v1 implementation doesn't already have a `list` operation. Moving this there would keep the v2 api space clean for now. Also, if you choose to not change the `operationId` below, then the generated client will create the method as `listTables`()`, which would be more clear that it's a v1 endpoint, not a v2 endpoint. However, I would recommend explicitly providing a unique operationId to make it clear this is a v1 endpoint.

emerkle826 (changes requested):

management-api-agent-3.x/src/main/java/com/datastax/mgmtapi/rpc/ObjectSerializer3x.java
olim7t: 👍 I missed it because the dse7 profile was not enabled in my IDE.

management-api-server/src/test/java/com/datastax/mgmtapi/BaseDockerIntegrationTest.java
emerkle826: I've captured this in https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/362

management-api-agent-3.x/src/test/java/com/datastax/mgmtapi/rpc/ObjectSerializer3xTest.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Please see the included license file for details.
+ */
+package com.datastax.mgmtapi.rpc;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ObjectSerializer3xTest extends ObjectSerializerTestBase<ObjectSerializer3x<Example>> {
```
emerkle826, line 10: Nice!

emerkle826 (approved):
Awesome!

emerkle826:
@olim7t I'll try to re-run the failing tests, none are indicative of an issue with this PR. When they complete, I think we are fine to merge this. My preference would be to do a squash and merge. And once it's merged, feel free to fire off the client build job so we can use this in Reaper. :+1:

PR #353, opened 2023-08-18
Integration branch: OpenAPI Client generation

emerkle826:
This is the aggregate PR to merge all of the smaller OpenAPI client generation updates into the main branch once it's all finished.


Miles-Garnsey:
One note on this PR for posterity: we were unable to provide adequate testing around the v2 repairs endpoint because (as discussed [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/389)) we could not test the generated client due to limitations with Maven.

Reviewers should note that (while we could technically test this using an HTTP-only testing approach) it seems late in the game to do so, and my understanding is that this is urgently needed to meet our next milestone.

adejanovski (approved):

emerkle826 (commented):
Just one nit suggestion. Not going to hold up the review for that.

PR #356, opened 2023-08-22
Fix DSE UBI Dockerfile so tests don't fail for new functionality

emerkle826:
I should create a ticket.....

This fixes the issue of tests failing against DSE UBI images by building the Management API bits from the branch the tests are running, as opposed to copying them from the most recently published DSE Ubuntu image (that may not have the current branch's functionality being tested)

**Fixes issue**
https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/357


Miles-Garnsey (approved):
Moved the merge target over to be the integration branch. All tests (including the previously failing DSE UBI tests) now appear to pass, so I'm approving. Thanks @emerkle826.

PR #358, opened 2023-08-22
Reaper endpoints: Async Repair Endpoint

Miles-Garnsey:
This PR modifies the existing `repair` method in `NodeOpsProvider` to accept arguments suitable for Reaper. When complete, it should also create a new v2 endpoint to accept new repairs and return a repairID which can later be queried for status updates.

**Fixes issue**
https://github.com/riptano/mission-control/issues/506


Miles-Garnsey:
@emerkle826 this is draft still, but I have a few questions and thought we might best collaborate with some concrete code to look at.

Questions:

1. Can the management API RPC mechanism handle polymporphic functions? We might need one which accepts a RingRange and one which accepts beginToken and endToken arguments, since you can repair non-contiguous ranges via the former. I know we've encountered issues with this in the past with the OpenAPI generator, so wanted to see what the best way to tackle it was.
2. Can you pass custom types into the RPC methods? The current methods seem to use only simple native Java types. Do they need to be registered as UDTs to make this work, or can you use Java types directly?
3. If we can't use polymorphism here to adapt to which signature we want to call, would it be reasonable to pass nulls instead?

The only place this matters is in the definition of the ranges, since the existing function just takes a start and end token, while the new function needs to take a list of start and end tokens (which is realised as a RingRange, but we could potentially use a List<Tuple> or List<List<BigInteger>> in the worst case).

emerkle826:
> 1. Can the management API RPC mechanism handle polymporphic functions? We might need one which accepts a RingRange and one which accepts beginToken and endToken arguments, since you can repair non-contiguous ranges via the former. I know we've encountered issues with this in the past with the OpenAPI generator, so wanted to see what the best way to tackle it was.

The RPC calls don't handle polymorphism (see [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-agent-common/src/main/java/com/datastax/mgmtapi/rpc/RpcObject.java#L37-L39)). RPC Methods have to have a unique RPC name, so we can't even register the same method with 2 different RPC names to handle a polymorphic situation.

emerkle826:
> 2\. Can you pass custom types into the RPC methods? The current methods seem to use only simple native Java types. Do they need to be registered as UDTs to make this work, or can you use Java types directly?

I think we can, but I'm not sure. To do so may require enhancements to some of the RPC serialization code.

emerkle826:
> 3\. If we can't use polymorphism here to adapt to which signature we want to call, would it be reasonable to pass nulls instead?

Yes, but I think in this case it might be better to have the API side of things determine which Range type is being requested and then build a specific RPC call for a RingRange or a start/end token. Unless I'm missing something, it seems like we will have 2 RPC methods on the Agent side, and the API/server side will handle the request and determine which of the RPC methods to call.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/models/RingRange.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Comparator;
+
+public final class RingRange {
+
+  public static final Comparator<RingRange> START_COMPARATOR
+      = (RingRange o1, RingRange o2) -> o1.start.compareTo(o2.start);
+
+  private final BigInteger start;
+  private final BigInteger end;
+
```
emerkle826, line 18: Something @olim7t just figured out is that model objects on the Agent side of things likely have to have their attributes as `public` in order for the Object serialization to work. See the code here:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/feature/get-tables/management-api-agent-3.x/src/main/java/com/datastax/mgmtapi/rpc/ObjectSerializer3x.java#L61

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -732,31 +735,67 @@ public void clearSnapshots(
...
-      repairSpec.put(RepairOption.INCREMENTAL_KEY, Boolean.toString(isIncremental));
-      if (isIncremental) {
-        // incremental repairs will fail if parallelism is not set
-        repairSpec.put(RepairOption.PARALLELISM_KEY, RepairParallelism.PARALLEL.getName());
-      }
-      ShimLoader.instance.get().getStorageService().repairAsync(keyspace, repairSpec);
+  }
+
+  // This version of repair() takes a list of RingRange objects instead of begin/end tokens.
+  @Rpc(name = "repair")
```
emerkle826, line 770: This won't work. We'll need a different Rpc `name` value or the registration will complain of duplicate names.

Miles-Garnsey:
> Yes, but I think in this case it might be better to have the API side of things determine which Range type is being requested and then build a specific RPC call for a RingRange or a start/end token. Unless I'm missing something, it seems like we will have 2 RPC methods on the Agent side, and the API/server side will handle the request and determine which of the RPC methods to call.

Gotcha. I've just been looking at this and I think the way to do it is actually as follows:

1. The current `repair()` method becomes a more general one which allows repair of multiple non-contiguous ranges via the RingRange<List> parameter (I'll fix the visibility of the fields as you suggest).
2. The new v2 Resource class will also accept a RingRange DTO.
3. The old endpoint will continue to accept beginToken and endToken, but will translate them into a single element List<RingRange> before passing off to the single `repair()` method.

Sound good?

Miles-Garnsey:
@emerkle826 can you give me a hand with this when you're in? I'm getting the below statement about illegally formatted files from CI:

```
Error:  To fix formatting errors, run "mvn com.coveo:fmt-maven-plugin:format"
Error:  Non complying file: /home/runner/work/management-api-for-apache-cassandra/management-api-for-apache-cassandra/management-api-agent-common/src/main/java/com/datastax/mgmtapi/rpc/models/RingRange.java
Error:  Non complying file: /home/runner/work/management-api-for-apache-cassandra/management-api-for-apache-cassandra/management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
Error:  Failed to execute goal com.coveo:fmt-maven-plugin:2.9:check (default) on project datastax-mgmtapi-agent-common: Found 2 non-complying files, failing build -> [Help 1]
Error:  
```

But when I run the suggested command it doesn't appear to change any files, or come up with any further information about what is wrong with the formatting. Do you know what I'm supposed to do with this?

emerkle826:
> @emerkle826 can you give me a hand with this when you're in? I'm getting the below statement about illegally formatted files from CI:
> 
> ```
> Error:  To fix formatting errors, run "mvn com.coveo:fmt-maven-plugin:format"
> Error:  Non complying file: /home/runner/work/management-api-for-apache-cassandra/management-api-for-apache-cassandra/management-api-agent-common/src/main/java/com/datastax/mgmtapi/rpc/models/RingRange.java
> Error:  Non complying file: /home/runner/work/management-api-for-apache-cassandra/management-api-for-apache-cassandra/management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
> Error:  Failed to execute goal com.coveo:fmt-maven-plugin:2.9:check (default) on project datastax-mgmtapi-agent-common: Found 2 non-complying files, failing build -> [Help 1]
> Error:  
> ```
> 
> But when I run the suggested command it doesn't appear to change any files, or come up with any further information about what is wrong with the formatting. Do you know what I'm supposed to do with this?

Just to close the loop here, you can run `mvn fmt:format` from the root of the project and it should auto format all of the source to the Google Java format standard. I have this documented way down in the README [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/#code-formatting). It covers Java source, license checks and XML formatting as well. Also, you will want to enable the DSE profiles too, or the formatting will skip those projects (i.e. `-P dse,dse7, trunk`).

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/RepairResources.java
```diff
@@ -0,0 +1,118 @@
...
+          Optional<RepairParallelism> repairParallelism =
+              request.repairParallelism == null
+                  ? Optional.empty()
+                  : Optional.of(request.repairParallelism);
+          Optional<Collection<String>> datacenters =
+              request.datacenters == null ? Optional.empty() : Optional.of(request.datacenters);
+          Optional<Integer> repairThreadCount =
+              request.repairThreadCount == null
+                  ? Optional.empty()
+                  : Optional.of(request.repairThreadCount);
```
emerkle826, line 94: Thought:
Here, you do all the `null` checks so you can convert them into `Optional.empty()`. I don't see the value in cluttering this all up. The benefit of using `Optional` would be to distinguish between a missing value and an empty value. But we don't check for `null` in NodeOpsProvider, and we don't really send `null` through the RPC. So there is no difference in just sending `null` here and moving the null checks into NodeOpsProvider.

emerkle826: Suggestion:
If we stay with `Optional` use here, then you can simply a lot of this
```java
          Optional<List<RingRange>> associatedTokens =
              request.associatedTokens == null
                  ? Optional.empty()
                  : Optional.of(request.associatedTokens);
```
with
```java
          Optional<List<RingRange>> associatedTokens =
              Optional.ofNullable(request.associatedTokens);
```

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/models/RepairRequestResponse.java
```diff
@@ -0,0 +1,36 @@
...
+  @JsonProperty(value = "repair_id", required = true)
+  public final String repairID;
+
+  @JsonCreator
+  public RepairRequestResponse(
+      @JsonProperty(value = "repair_id", required = true) String repairID) {
+    this.repairID = repairID;
+  }
+
+  public boolean equals(Object o) {
```
emerkle826, line 23: Issue:
Add `@Override`

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/models/RepairRequestResponse.java
```diff
@@ -0,0 +1,36 @@
...
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    return Objects.equals(repairID, ((RepairRequestResponse) o).repairID);
+  }
+
+  public int hashCode() {
```
emerkle826, line 33: Issue:
Add `@Override`

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/models/RepairRequest.java
```diff
@@ -0,0 +1,86 @@
...
+    return Objects.equals(keyspace, ((RepairRequest) o).keyspace)
+        && Objects.equals(tables, ((RepairRequest) o).tables)
+        && Objects.equals(fullRepair, ((RepairRequest) o).fullRepair)
+        && Objects.equals(associatedTokens, ((RepairRequest) o).associatedTokens)
+        && Objects.equals(datacenters, ((RepairRequest) o).datacenters)
+        && Objects.equals(repairParallelism, ((RepairRequest) o).repairParallelism)
+        && Objects.equals(repairThreadCount, ((RepairRequest) o).repairThreadCount);
+  }
+
+  public int hashCode() {
```
emerkle826, line 77: Issue:
Add `@Override`

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/models/RepairRequest.java
```diff
@@ -0,0 +1,86 @@
...
+    this.tables = tables;
+    this.fullRepair = fullRepair;
+    this.notifications = notifications;
+    this.associatedTokens = associatedTokens;
+    this.datacenters = datacenters;
+    this.repairParallelism = repairParallelism;
+    this.repairThreadCount = repairThreadCount;
+  }
+
+  public boolean equals(Object o) {
```
emerkle826, line 61: Issue:
Add `@Override`

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/models/RepairParallelism.java
```diff
@@ -0,0 +1,38 @@
...
+  private RepairParallelism(String name) {
+    this.name = name;
+  }
+
+  @JsonValue
+  public String getName() {
+    return this.name;
+  }
+
+  public String toString() {
```
emerkle826, line 35: Issue:
Add `@Override`

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/models/RingRange.java
```diff
@@ -0,0 +1,42 @@
...
+    end = new BigInteger(range[1]);
+  }
+
+  public BigInteger getStart() {
+    return start;
+  }
+
+  public BigInteger getEnd() {
+    return end;
+  }
```
emerkle826, line 41: Suggestion:
Would probably be a good idea to add `hashCode()` and `equals()`

management-api-server/src/test/java/com/datastax/mgmtapi/K8OperatorResourcesTest.java
```diff
@@ -2111,4 +2108,16 @@ public void testMoveAsync_MissingNewToken() throws Exception {
...
+
+    Dispatcher dispatcher;
+    CqlService cqlService;
+
+    MockHttpResponse invoke(HttpRequest request) {
+      MockHttpResponse response = new MockHttpResponse();
+      dispatcher.invoke(request, response);
+      return response;
+    }
+  }
```
emerkle826, line 2122: Issue:
No need to move the inner class definition down. It just clutters up the diff

emerkle826: I just realized this might be due to IDE settings. You can refer to [this](https://github.com/k8ssandra/management-api-for-apache-cassandra/#using-checkstyle-in-an-ide) for the Googe Java format style

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/RepairResources.java
```diff
@@ -0,0 +1,118 @@
...
+
+@Path("/api/v2/repairs")
+public class RepairResources extends BaseResources {
+
+  public RepairResources(ManagementApplication application) {
+    super(application);
+  }
+
+  @PUT
+  @Operation(summary = "Initiate a new repair", operationId = "v2PutRepair")
```
emerkle826, line 38: Suggestion:
```suggestion
  @Operation(summary = "Initiate a new repair", operationId = "putRepairV2")
```
The `operationId` comes mostly into play when the client is generated. For `getTables()`, we've gone with the template of `<operation>V#`. For all of the v0 and v1 endpoints, the method names are all over the place, so we don't exclusively use the HTTP verbs (GET, PUT, DELETE, etc).

So for v2, I think I'd like to append the version to the end. My reasoning is that when trying to find a method you want to use in a client application (i.e. Reaper), you are more likely going to look for the function you want before looking for the version of it. IDEs tend to give you pop-ups with all the available methods on the generated client, and they are usually in alphabetical order. So having the `v2` prepended to the beginning of the method name will mean that you need to scroll down to find the v2 version specifically. I think it's better to have, for example:
- getKeyspaces()
- getKeyspacesV1()
- getKeyspacesV2()

Having said all that, and going back to my earlier point that we don't consistently use the approrpiate HTTP verbs as the method prefix, this nice neat grouping of similar methods isn't going to happen anyway. Still, I think it would be easier to have the version at teh end of the method name as opposed to prefixing it.

emerkle826 (changes requested):
I know this is still a WIP, but I've given it an initial review for now.

emerkle826 (changes requested):

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
```diff
@@ -503,11 +504,16 @@ public Response repair(RepairRequest repairRequest) {
           }
           app.cqlService.executePreparedStatement(
               app.dbUnixSocketFile,
-              "CALL NodeOps.repair(?, ?, ?, ?)",
+              "CALL NodeOps.repair(?, ?, ?, ?, ?, ?, ?)",
```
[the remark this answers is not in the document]
emerkle826: I just pushed a fix for this

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -739,96 +742,101 @@ public String repair(
...
-
-      // handle incremental vs full
-      boolean isIncremental = Boolean.FALSE.equals(full);
-      repairSpec.put(RepairOption.INCREMENTAL_KEY, Boolean.toString(isIncremental));
-      if (isIncremental) {
-        // incremental repairs will fail if parallelism is not set
-        repairSpec.put(RepairOption.PARALLELISM_KEY, RepairParallelism.PARALLEL.getName());
-      }
+    assert (keyspace != null);
+    Map<String, String> options = new HashMap<>();
```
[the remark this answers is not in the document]
Miles-Garnsey: If you prefer it I'm happy to make this change.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -739,96 +742,101 @@ public String repair(
...
+    repairParallelism.map(rPar -> options.put(RepairOption.PARALLELISM_KEY, rPar.getName()));
+    options.put(RepairOption.INCREMENTAL_KEY, Boolean.toString(!full));
+    repairThreadCount.map(
+        tCount ->
+            options.put(RepairOption.JOB_THREADS_KEY, Integer.toString(tCount == 0 ? 1 : tCount)));
+    options.put(RepairOption.TRACE_KEY, Boolean.toString(Boolean.FALSE));
+    options.put(RepairOption.COLUMNFAMILIES_KEY, StringUtils.join(tables, ","));
```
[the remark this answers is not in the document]
Miles-Garnsey: This is a good question and something I was discussing with the team last night. I think if the argument is optional I'd rather wrap it in an Optional<> in this part of the codebase. What do you think? I didn't previously do this to avoid changing the existing handling, but as I've already done that, it seems like maybe I should just bite the bullet and use Optional<>...

Then yes, you have the other two cases where the List is empty or the argument itself is null. Dealing with all of these cases is honestly quite a headache as I'm not sure how things get deserialised from json!

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -739,96 +742,101 @@ public String repair(
...
+    if (full) {
+      associatedTokens.map(
```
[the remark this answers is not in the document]
Miles-Garnsey: I think input sanitization should be handled in the Resources layer, no? By the time it gets to the NodeOpsProvider I'd rather have everything in Optionals, but tell me if you disagree. I do wonder if introducing Optionals here just adds more complexity as now we have a third type of nullity.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
```diff
@@ -503,11 +504,16 @@ public Response repair(RepairRequest repairRequest) {
...
               repairRequest.keyspaceName,
               repairRequest.tables,
               repairRequest.full,
-              false);
+              // The default repair does not allow for specifying things like parallelism,
+              // threadCounts, source DCs or ranges etc.
+              Optional.empty(),
+              Optional.empty(),
+              Optional.empty(),
+              Optional.empty());
```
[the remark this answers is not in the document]
Miles-Garnsey: See my comment [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/358#discussion_r1305002630), I was just thinking on the same topic and had basically the same thoughts. 

Maybe we can jump on a call to discuss if you have time at some stage.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/models/RepairRequest.java
Miles-Garnsey: Oop, I'm sorry, I'm a hypocrite :(

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/models/RingRange.java
Miles-Garnsey: Thanks, done.

management-api-server/src/test/java/com/datastax/mgmtapi/K8OperatorResourcesTest.java
Miles-Garnsey: I've installed the google style plugin, hopefully everything is formatted better now.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/RepairResources.java
Miles-Garnsey: OK, I'm happy to go along with this (PS: remember to use issue/suggesiton.

Miles-Garnsey:
@olim7t can Erik and I get some feedback from you on the topic of Optionals? The context is that in the NodeOpsProvider, we want to avoid adding keys to the repair options map under several circumstances:

1. When the option is passed in as `null`.
2. When the option is an empty collection.
3. When the option is the zero value of the type.

This is because we want the NodeOpsProvider logic to service v0-v3 endpoints, and some options are unavailable in the v0/v1 endpoints. 

To facilitate this, I have the new options being passed as `Optional<>` right now, and in the resources layers we will munge the data to retain existing logic and pass empty `Optionals` when the endpoint doesn't support those options. 

But Erik has rightly noted that the calls over the RPC channel aren't type checked, so wrapping things in `Optional<>` may not be hugely valuable. He would prefer to keep null checks inside the `NodeOpsProvider` and consolidate emptiness  into null values in the resources layer. 

Neither of us have a strong feeling on this - do you have any thoughts?

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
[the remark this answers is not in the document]
Miles-Garnsey: I think we've discussed this. We will remove the Optionals<> depending on Olivier's feedback. For now I'll keep them and expand the input sanitation.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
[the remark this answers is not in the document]
Miles-Garnsey: I think we've discussed this. We will remove the Optionals<> depending on Olivier's feedback. For now I'll keep them and expand the input sanitation.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/RepairResources.java
Miles-Garnsey: I will implement this pending Olivier's feedback.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -737,98 +740,106 @@ public void clearSnapshots(
   @Rpc(name = "repair")
   public String repair(
       @RpcParam(name = "keyspaceName") String keyspace,
-      @RpcParam(name = "tables") List<String> tables,
+      @RpcParam(name = "tables") Optional<List<String>> tables,
       @RpcParam(name = "full") Boolean full,
-      @RpcParam(name = "notifications") boolean notifications)
+      @RpcParam(name = "notifications") boolean notifications,
+      @RpcParam(name = "repairParallelism") Optional<RepairParallelism> repairParallelism,
+      @RpcParam(name = "datacenters") Optional<Collection<String>> datacenters,
```
emerkle826, line 747: Issue:
`Collection` isn't handled by the serializer either.

emerkle826 (changes requested):

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -737,98 +740,106 @@ public void clearSnapshots(
   @Rpc(name = "repair")
   public String repair(
       @RpcParam(name = "keyspaceName") String keyspace,
-      @RpcParam(name = "tables") List<String> tables,
+      @RpcParam(name = "tables") Optional<List<String>> tables,
       @RpcParam(name = "full") Boolean full,
-      @RpcParam(name = "notifications") boolean notifications)
+      @RpcParam(name = "notifications") boolean notifications,
+      @RpcParam(name = "repairParallelism") Optional<RepairParallelism> repairParallelism,
```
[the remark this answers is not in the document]
Miles-Garnsey: > The second issue is more of a confusing thing. The v2 model object looks a lot like the Cassandra object, but they are not the same. And in the Resource code, we are not sending the Cassandra object, we are sending the V2 model object. So this will never deserialize correctly since we are expecting the Cassandra object in this method.

Woops... Sorry, I should have spotted that one :( 

Wouldn't we be better off adding a generic serializer for the collections interface to make things more generic? I have to be honest, I do not like the way management API is not using the type system very much throughout. I understand some of that is unavoidable due to the way the RPC calls work, but if we can't even serialise our own types, that is unfortunate.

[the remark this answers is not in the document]
Miles-Garnsey: Actually, I need to understand why we aren't just serializing straight to json or using some off the shelf serialization here - was there a reason for writing our own serializer?

Miles-Garnsey:
> I haven't checked every implementation, but looking at 3.x's RepairOption.parse, it looks pretty safe. It's just looking up known options in the map, if there are a few extra unknown ones, they should be ignored.

It isn't so much about unknown options, it is more about setting the option values to null or empty collections. @olim7t

emerkle826 (approved):
Ready for merge

Miles-Garnsey:
This fixed #405

PR #365, opened 2023-08-25
Add operation to get active compactions (fixes #364)

### LINKED ISSUE #364: Add operation go get active compactions

olim7t:
Fixes #364

### DIFF
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/Compaction.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/Compaction.java
@@ -1,0 +1,191 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Please see the included license file for details.
+ */
+package com.datastax.mgmtapi.resources.models;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Describes the state of an active compaction running on the server.
+ *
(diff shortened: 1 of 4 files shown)


Miles-Garnsey:
@olim7t there seem to be a lot of failing tests on this one. Is it a draft or ready for review?

emerkle826:
@olim7t @Miles-Garnsey Sorry I took so long on the review on this one. The failing tests seem to be the typical flaky `refCnt` failures, plus a few weird `docker not found` things. I'm re-running the failed tests now.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/Compaction.java
```diff
@@ -0,0 +1,191 @@
...
+  private static final String COMPLETED_KEY = "completed";
+  private static final String TOTAL_KEY = "total";
+  private static final String TASK_TYPE_KEY = "taskType";
+  private static final String UNIT_KEY = "unit";
+  private static final String COMPACTION_ID_KEY = "compactionId";
+  private static final String SSTABLES_KEY = "sstables";
+  private static final String TARGET_DIRECTORY_KEY = "targetDirectory";
+  private static final String OPERATION_TYPE_KEY = "operationType";
+  private static final String OPERATION_ID_KEY = "operationId";
+  private static final String DESCRIPTION_KEY = "description";
```
emerkle826, line 37: Thought:
This is fine. This approach is better than introducing Cassandra/DSE dependencies here as that would force us to break the server/API side of the code up into version specific projects like we do for the Agents.

emerkle826 (approved):
This looks good to me. It would be nice to have a unit or integration test, but the nature of this makes that tricky. A unit test wouldn't really test much with mocks because we'd be mocking the return of a map of key-value pairs. I might create a ticket to try to come up with an integration test for it in the future, but I'm not blocking this PR to wait for that.

PR #366, opened 2023-08-28
Repair endpoint: Remove Optional and other objects not handled by RPC serialization

Miles-Garnsey:
Raising a PR so we can discuss the changes more easily.


management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -740,39 +737,46 @@ public void clearSnapshots(
...
       @RpcParam(name = "keyspaceName") String keyspace,
-      @RpcParam(name = "tables") Optional<List<String>> tables,
+      @RpcParam(name = "tables") List<String> tables,
       @RpcParam(name = "full") Boolean full,
       @RpcParam(name = "notifications") boolean notifications,
-      @RpcParam(name = "repairParallelism") Optional<RepairParallelism> repairParallelism,
-      @RpcParam(name = "datacenters") Optional<Collection<String>> datacenters,
-      @RpcParam(name = "associatedTokens") Optional<List<RingRange>> associatedTokens,
-      @RpcParam(name = "repairThreadCount") Optional<Integer> repairThreadCount)
+      @RpcParam(name = "repairParallelism") String repairParallelism,
```
Miles-Garnsey, line 743: Question: Is it not possible to use an enum here? We could probably source the enum from Cassandra to avoid direct dependancies between the agent and server.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -740,39 +737,46 @@ public void clearSnapshots(
...
+      @RpcParam(name = "datacenters") List<String> datacenters,
+      @RpcParam(name = "associatedTokens") String ringRangeString,
```
Miles-Garnsey, line 745: Question: There's no ability to serialise any non-primitive types over the RPC channel?

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -740,39 +737,46 @@ public void clearSnapshots(
...
+      @RpcParam(name = "repairThreadCount") int repairThreadCount)
```
Miles-Garnsey, line 746: Question: Slightly concerned that we can't distinguish a zero from a null here? Would it be better to use an Integer? Or can we not serialise that?

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -740,39 +737,46 @@ public void clearSnapshots(
...
+          && !RepairParallelism.PARALLEL.getName().equals(repairParallelism)) {
+        throw new IOException(
+            "Invalid repair combination. Incremental repair if Parallelism is not set");
+      }
+      // Incremental repair and parallelism should be set
+      repairSpec.put(RepairOption.PARALLELISM_KEY, RepairParallelism.PARALLEL.getName());
+    }
+    // thread count should be at least 1
+    repairSpec.put(
+        RepairOption.JOB_THREADS_KEY, Integer.toString(Integer.max(1, repairThreadCount)));
```
Miles-Garnsey, line 770: Question: Do we need to set this explicitly? I thought leaving it unset was best if the original request didn't contain the key.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -740,39 +737,46 @@ public void clearSnapshots(
...
-                  StringUtils.join(
-                      aTokens.stream()
-                          .map(token -> token.getStart() + ":" + token.getEnd())
-                          .collect(Collectors.toList()),
-                      ",")));
-    }
-    datacenters.map(
-        dcs -> repairSpec.put(RepairOption.DATACENTERS_KEY, StringUtils.join(dcs, ",")));
+    // add ranges if reapir is full
+    if (full && ringRangeString != null && !ringRangeString.isEmpty()) {
```
Miles-Garnsey, line 773: Suggestion: I'd be more inclined to throw an error here if `!full && ringRangeString != null`, since it indicates an invalid request AFAICT?

Miles-Garnsey (commented):
I have a bunch of questions on this one, happy to approve for merge into my branch once they're answered. I don't really know how the serialisation stuff works (although I've had a read through as best I can) so that's probably the source of a lot of my questions.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
emerkle826: > Question: Is it not possible to use an enum here? We could probably source the enum from Cassandra to avoid direct dependencies between the agent and server.

It is possible, however it would have to be defined in the `management-api-agent-common` or `management-api-common` projects and then also referenced in the `management-api-server` project. It might also have to be added to all of the flavors of `GenericSerializerXYZ` in order for it to be serialized/deserialized correctly.

Using 3.11 to demonstrate, if you look at the RpcMethod stuff here:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-agent-3.x/src/main/java/com/datastax/mgmtapi/rpc/RpcMethod3x.java#L79-L82
you will see where the serializers for any RPC method arguments are setup.

A little more complicated looking, but the same basic idea for serializers for returned objects from RPC methods is here:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-agent-3.x/src/main/java/com/datastax/mgmtapi/rpc/RpcMethod3x.java#L91-L104

Basically, RPC method arguments will be handled by GenericSerializer, and RPC methods return values are handled by ObjectSerializer.

Regarding RPC method arguments, we can look at the 3.11 GenericSerializer code:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-agent-3.x/src/main/java/com/datastax/mgmtapi/rpc/GenericSerializer3x.java#L33-L67
and
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-agent-3.x/src/main/java/com/datastax/mgmtapi/rpc/GenericSerializer3x.java#L102-L142
You'll see that it basically has to map Cassandra types to/from Java types, and it also handles Java Lists, Sets and Maps. If we want to handle an enum, it is going to have to be added in here so that the argument can be serialized/deserialzed correctly for the RPC call. This is the slippery slope I mentioned in the other PR. We could keep adding and adding and adding and it will get a bit unruly. Looking at the code, it seems this was intended to handle basic types, with the small addition of Lists, Sets and Maps.

You might have the desire to then use the Cassandra type `RepairParallelism` here, since it is an enum and already defined in Cassandra. There is a big potential problem with that though. The problem is that anything defined in Cassandra potentially needs to be compiled against Cassandra specific jarfiles. This is why we have so many agent projects. The code is mostly the same (not always), but if it's compiled against the wrong version of Cassandra jarfiles, the agents produced fail when deployed in Cassandra instances of different versions than what it was compiled against. We currently do NOT have a dependency upon Cassandra jarfiles in the management-api-server project, and we need to keep it that way. If we introduce Cassandra dependnecies there, then we will have just as many server projects as we have agent projects.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
emerkle826: there is, but we have to explicitly handle any ones we want. see https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/366#discussion_r1307815316

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
emerkle826: For thread count, we could use an Integer object, but then it adds a null-check in at least the NodeOpsProvider. With it being a primitive, the caller will fail if an `int` isn't sent over the wire. And the logic in NodeOpsProvider is:
```java
RepairOption.JOB_THREADS_KEY, Integer.toString(Integer.max(1, repairThreadCount))
```
so anything less than `1` becomes `1`, so the logic would simply set `0` and `null` to a `1` anyway.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
emerkle826: I guess we could then. But the logic then becomes something like:
```java
if (repairThreadCount > 1) {
    repairSpec.put(RepairOption.JOB_THREADS_KEY, Integer.toString(repairThreadCount));
}
```
So it's only added if the request asked for more than 1 thread.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
Miles-Garnsey: OK, got it. Sounds like a string is going to be the most efficient way to do this 😫

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
Miles-Garnsey: I think that's the right way, yep.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
Miles-Garnsey: The thing is, I think we want to leave this unset if it is null, so indeed there should be a null check and we shouldn't add it to the options map.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
Miles-Garnsey: I'm not sure! Based on your logic here it seemed that you only wanted to add the `RANGES_KEY` when the repair type was `full`. So I assumed that ranges was only valid for full repairs. 

In my original logic, I think we just don't add the ranges if they aren't specified, which might be the the best way to do it. Let Cassandra throw an error if there's an error to be thrown?

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
emerkle826: OK, I will update it then to do it like that.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
emerkle826: Ok, I'll investigate this and update accordingly.

Miles-Garnsey: Awesome!

Miles-Garnsey:
I'm going to merge this, the only area where we need to discuss further is:

```
repairSpec.put(RepairOption.JOB_THREADS_KEY, repairThreadCount.toString());
```

Where I think `repairThreadCount` I think should be an `Integer`, and we should leave unset if null or 0. I'll make that change in my branch once merged, and @emerkle826 can then confirm he's happy with it via review there.

PR #368, opened 2023-08-29
Cancel repairs endpoint

Miles-Garnsey:
This PR solves the other half of this [ticket](https://github.com/riptano/mission-control/issues/506).

### DIFF
--- a/management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
+++ b/management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
@@ -853,6 +853,11 @@ public class NodeOpsProvider {
     return job.getJobId();
   }
 
+  @Rpc(name = "stopAllRepairs")
+  public void stopAllRepairs() {
+    ShimLoader.instance.get().getStorageService().forceTerminateAllRepairSessions();
+  }
+
   @Rpc(name = "move")
   public String move(
       @RpcParam(name = "newToken") String newToken, @RpcParam(name = "async") boolean async)
--- a/management-api-server/doc/openapi.json
+++ b/management-api-server/doc/openapi.json
@@ -1678,6 +1678,23 @@
       }
     },
     "/api/v2/repairs" : {
+      "delete" : {
+        "operationId" : "deleteRepairsV2",
+        "responses" : {
+          "202" : {
+            "content" : {
+              "application/json" : {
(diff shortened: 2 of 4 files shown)


management-api-server/src/main/java/com/datastax/mgmtapi/resources/common/BaseResources.java
```diff
@@ -65,6 +65,7 @@ protected Response handle(Callable<Response> action) {
           .entity("Internal connection to Cassandra closed")
           .build();
     } catch (Throwable t) {
+      t.printStackTrace();
```
emerkle826, line 68: Thought:
I put this in to help debug some unexpected 500 responses. May want to convert it to a log? Or simply get rid of it?

management-api-server/src/test/java/com/datastax/mgmtapi/BaseDockerIntegrationTest.java
```diff
@@ -59,7 +59,7 @@ protected void failed(Throwable e, Description description) {
           System.err.flush();
 
           if (null != docker) {
-            int numberOfLines = 100;
+            int numberOfLines = 1000;
```
emerkle826, line 62: thought:
I upped this here as it's the number of lines of the Cassandra logfile dumped out during a failed IT test. Sometimes the error is more than 100 lines from the end of the logfile. Since this is only for tests, I think it's ok to up this.

emerkle826 (approved):
This looks good, just need to sync with #366 changes once they are settled.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/common/BaseResources.java
Miles-Garnsey: Why not leave it, if it helps debugging it may as well stay.

management-api-server/src/test/java/com/datastax/mgmtapi/BaseDockerIntegrationTest.java
Miles-Garnsey: I'm OK with this too :)

emerkle826:
@Miles-Garnsey I pushed a commit to fix the formatting issue and duplicate methods. Also, you were missing the change to `openapi.json`, so I added that to the commit as well. Everything looks good now. :+1:

PR #370, opened 2023-08-31
 Replace use of BigIntegers in Resources' model classes with Longs

Miles-Garnsey:
Replace usage of BigIntegers with Longs. Hopefully this will fix the issues with BigIntegers being represented in open API spec as ints.

### DIFF
--- a/management-api-server/doc/openapi.json
+++ b/management-api-server/doc/openapi.json
@@ -2147,10 +2147,12 @@
         "type" : "object",
         "properties" : {
           "end" : {
-            "type" : "integer"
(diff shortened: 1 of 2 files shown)


emerkle826 (approved):
I looked into the OpenAPI client generation and I think this is the best solution. We could add typeMappings to the generator to get `BigInteger` in the client implementation, but it would convert either all `int` or `long` Java primitives to `BigInteger`. Also, it seems that Java's `long` is 64-bit, which covers Cassandra's token range values, so I think we are good here.

PR #376, opened 2023-09-05
Add hot reloading of SslContext

### LINKED ISSUE #375: Add SSL/TLS hot reloading and allow TLS v1.3
Currently, the only accepted version of TLS is set to v1.2, although Netty set v1.3 in the default settings already ~2 years ago. Lets remove this restriction and allow connections with 1.3 also.

The other issue is that SslContext is only created on server startup, so it is never refreshed and the certs are never reloaded. We need something to automatically reload them whenever a change happens (as cass-operator would always load from the Secret the latest one - so should we if the secret on the disk has been modified).

burmanm:
Create a Filewatcher to detect any changes in the filesystem related to these TLS files (key, CA, cert). If something happens on the disk to these files, simply reload the SslContext.

Also, change v1.2 restriction to Netty defaults (including support for TLS v1.3)

Fixes #375

### DIFF
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/Cli.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/Cli.java
@@ -513,6 +513,7 @@ public class Cli implements Runnable {
                 break;
               }
               if (reloadNeeded) {
+                logger.info("Detected change in the SSL/TLS certificates, reloading.");
                 createSSLContext();
               }
             } catch (InterruptedException e) {
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/Cli.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/Cli.java
@@ -452,7 +452,8 @@ public class Cli implements Runnable {
     checkUnixSocket();
   }
 
-  private void createSSLContext() throws SSLException {
+  @VisibleForTesting
+  void createSSLContext() throws SSLException {
     this.sslContext =
         SslContextBuilder.forServer(tlsCert, tlsKey)
             .trustManager(tlsCaCert)
@@ -461,7 +462,8 @@ public class Cli implements Runnable {
             .build();
   }
 
-  private void createSSLWatcher() throws IOException {
+  @VisibleForTesting
+  void createSSLWatcher() throws IOException {
     // Watch for tls_cert_file, tls_key_file and tls_ca_cert_file, add all their directories to
     // Filesystem Watcher
     WatchService watchService = FileSystems.getDefault().newWatchService();
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/Cli.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/Cli.java
@@ -46,15 +46,21 @@ import java.io.File;
 import java.io.IOException;
 import java.net.SocketAddress;
 import java.net.URI;
+import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Paths;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
@@ -74,7 +80,6 @@ import org.slf4j.LoggerFactory;
     name = "cassandra-management-api",
     description = "REST service for managing an Apache Cassandra or DSE node")
 public class Cli implements Runnable {
-  public static final String PROTOCOL_TLS_V1_2 = "TLSv1.2";
 
   static {
     InternalLoggerFactory.setDefaultFactory(new Slf4JLoggerFactory());
@@ -136,6 +141,8 @@ public class Cli implements Runnable {
       description = "Path to trust certs signed only by this CA")
   private String tls_ca_cert_file;
 
+  private File tlsCaCert;
+
   @Path(writable = false)
   @Option(
       name = {"--tlscert"},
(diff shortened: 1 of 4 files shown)


management-api-server/src/main/java/com/datastax/mgmtapi/Cli.java
```diff
@@ -436,18 +452,85 @@ void preflightChecks() {
...
+        StandardWatchEventKinds.ENTRY_CREATE,
+        StandardWatchEventKinds.ENTRY_DELETE,
+        StandardWatchEventKinds.ENTRY_MODIFY);
+    tlsCaCertParent.register(
+        watchService,
+        StandardWatchEventKinds.ENTRY_CREATE,
+        StandardWatchEventKinds.ENTRY_DELETE,
+        StandardWatchEventKinds.ENTRY_MODIFY);
+
+    ExecutorService executorService = Executors.newSingleThreadExecutor();
```
Miles-Garnsey, line 490: Suggestion: Can we piggy back off any existing thread pool? I'm a little hesitant to create random single threaded executors this deep down in the application. 

Maybe it would be good to inject the executor from the `main` function in either event.

management-api-server/src/main/java/com/datastax/mgmtapi/Cli.java
```diff
@@ -436,18 +452,85 @@ void preflightChecks() {
...
+                }
+              }
+              if (!key.reset()) {
+                // The watched directories have disappeared..
+                break;
+              }
+              if (reloadNeeded) {
+                createSSLContext();
+              }
+            } catch (InterruptedException e) {
```
Miles-Garnsey, line 518: Suggestion: Won't it only be interrupted as part of application shut down? Maybe no need to log here.

Miles-Garnsey (commented):
I've left a few preliminary comments, I think this is one I'd like to test manually. Have you done so yourself?

Do you have any thoughts/scripts from when you did so? My thinking is that I probably need to create some certs, add them to truststores/keystores and then swap them for a new set, while checking that the client's old certs are now rejected and the new ones start working.

management-api-server/src/main/java/com/datastax/mgmtapi/Cli.java
burmanm: Threads can get interruptedexception without shutting down. It shouldn't happen, but if it does then that's an indication of something else.

management-api-server/src/main/java/com/datastax/mgmtapi/Cli.java
burmanm: No, since this will block that executioner, so if we piggyback some existing ones, we need to ensure there's an existing single thread there always reserved for this one.

Miles-Garnsey: Ok fair, so I'll amend this suggestion to be: can we find where any other thread pools are created and ensure that any thread pools we're creating are created in the same place.

management-api-server/src/main/java/com/datastax/mgmtapi/Cli.java
Miles-Garnsey: Can we distinguish between the two cases and only log in the case where the InterruptedException does not relate to application shutdown? I don't like the idea of producing errors in the logs for normal application behaviour.

burmanm: It is never normal application behavior to get InterruptedException.

burmanm:
No need to test this tomorrow, I need to verify one thing that felt suspicious today in my testing.

management-api-server/src/test/java/com/datastax/mgmtapi/NettyTlsClientAuthTest.java
```diff
@@ -391,4 +396,56 @@ public void testManagementAPIWithTLS() throws IOException {
...
+    new File(cassSock).deleteOnExit();
+
+    Path tempDirectory = Files.createTempDirectory("reload-test");
+
+    List<String> extraArgs =
+        IntegrationTestUtils.getExtraArgs(
+            NettyTlsClientAuthTest.class, "", temporaryFolder.getRoot());
+
+    File trustCertFile =
+        IntegrationTestUtils.getFile(getClass(), "mutual_auth_client_cert_chain.pem");
```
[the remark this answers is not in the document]
burmanm: Yes.

burmanm:
Alright, that was nothing - the operator does indeed create new connection each time, so the new SSL gets loaded. It takes around 5 minutes for Kubernetes to detect that. That is, Netty reruns the:

```java
channelPipeline.addFirst(sslContext.newHandler(ch.alloc()));
```

On each new connection and our reference to the sslContext has now been updated when the cert gets refreshed. 

As there's no logging in this current PR, I know it's a bit tricky to see when it actually does reload something. I added logging to my personal fork, but I'm happy to add here to notify that SSLs have changed. And from that, the process I used to test this manually:

cass-operator has this test, which loads MTLS certs to management-api and verifies it works: https://github.com/k8ssandra/cass-operator/blob/master/tests/test_mtls_mgmt_api/test_mtls_mgmt_api_suite_test.go

The parts in it are what I used. So, I created my own image:

```
docker buildx build -t michaelburman290/cass-management-api:4.1.2 -f Dockerfile-4_1.ubi8 . --load
kind load docker-image michaelburman290/cass-management-api:4.1.2
```

And then I modified ``tests/testdata/oss-one-node-dc-with-mtls.yaml`` with the following patch:

```
diff --git a/tests/testdata/oss-one-node-dc-with-mtls.yaml b/tests/testdata/oss-one-node-dc-with-mtls.yaml
index 144c760..92a8051 100644
--- a/tests/testdata/oss-one-node-dc-with-mtls.yaml
+++ b/tests/testdata/oss-one-node-dc-with-mtls.yaml
@@ -5,7 +5,8 @@ metadata:
 spec:
   clusterName: cluster1
   serverType: cassandra
-  serverVersion: "3.11.7"
+  serverVersion: "4.1.2"
+  serverImage: michaelburman290/cass-management-api:4.1.2
   managementApiAuth:
     manual:
       clientSecretName: mgmt-api-client-credentials
@@ -22,6 +23,6 @@ spec:
   racks:
     - name: r1
   config:
-    jvm-options:
+    jvm-server-options:
       initial_heap_size: "512m"
       max_heap_size: "512m"
```

Then, deploy everything..

```
kubectl apply -f tests/testdata/mtls-certs-server.yaml
kubectl apply -f tests/testdata/mtls-certs-client.yaml
kubectl apply -f tests/testdata/oss-one-node-dc-with-mtls.yaml
```

Verify that they work. I then simply modified the mtls-certs-server.yaml with new stuff (using kubectl edit), another cert pairs and checked that the behavior has changed after the reloads have happened. I verified the reloads and new data by going to the pod and checking the filesystem and comparing the logs. After that, let cass-operator try to connect to it.. "whoops", it will fail since it tries to use the old certs-client.yaml stuff and that would no longer match whatever I had in my new ones.

emerkle826 (approved):
@Miles-Garnsey I've tested this locally and it works. Approving.

PR #377, opened 2023-09-07
Add Token Range to Endpoints endpoint

### LINKED ISSUE #374: Add token range to endpoint mapping
Clients need to get the token range to endpoint mapping, which is already provided by Cassandra. Expose and endpoint that allows for fetching this information.

```[tasklist]
### DoD
- [ ] An endpoint exists that exposes token range to endpoint mappings
```

emerkle826:
Fixes #374

### DIFF
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/resources/common/BaseResources.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/resources/common/BaseResources.java
@@ -8,6 +8,8 @@ package com.datastax.mgmtapi.resources.common;
 import com.datastax.mgmtapi.ManagementApplication;
 import com.datastax.mgmtapi.resources.helpers.ResponseTools;
 import com.datastax.oss.driver.api.core.NoNodeAvailableException;
+import com.datastax.oss.driver.api.core.cql.ResultSet;
+import com.datastax.oss.driver.api.core.cql.Row;
 import java.util.concurrent.Callable;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.core.Response;
@@ -70,5 +72,22 @@ public abstract class BaseResources {
           .entity(t.getLocalizedMessage())
           .build();
     }
+  }
+
+  protected boolean keyspaceExists(String keyspaceName) {
+    if (keyspaceName != null) {
+      try {
+        ResultSet result =
+            app.cqlService.executePreparedStatement(
+                app.dbUnixSocketFile, "CALL NodeOps.getKeyspaces()");
+        Row row = result.one();
+        if (row != null) {
+          return row.getList(0, String.class).contains(keyspaceName);
+        }
+      } catch (Exception e) {
+        // cxonnection issue or some other problem. Eat it and return false.
+      }
+    }
+    return false;
   }
 }
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/resources/common/BaseResources.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/resources/common/BaseResources.java
@@ -74,18 +74,22 @@ public abstract class BaseResources {
     }
   }
 
-  protected boolean keyspaceExists(String keyspaceName) {
+  /**
+   * Returns true if the specified keyspaceName is not null and a keyspace with the name exists.
+   * Returns false if the keyspaceName is null or if no keyspace with the name exists. Throws a
+   * ConnectionClosedException if there is an issue executing the RPC call to the Cassandra agent.
+   *
+   * @param keyspaceName The name of a keyspace you are looking for.
+   * @return True if the keyspace is found, false otherwise.
+   */
+  protected boolean keyspaceExists(String keyspaceName) throws ConnectionClosedException {
     if (keyspaceName != null) {
-      try {
-        ResultSet result =
-            app.cqlService.executePreparedStatement(
-                app.dbUnixSocketFile, "CALL NodeOps.getKeyspaces()");
-        Row row = result.one();
-        if (row != null) {
-          return row.getList(0, String.class).contains(keyspaceName);
-        }
-      } catch (Exception e) {
-        // cxonnection issue or some other problem. Eat it and return false.
+      ResultSet result =
+          app.cqlService.executePreparedStatement(
+              app.dbUnixSocketFile, "CALL NodeOps.getKeyspaces()");
+      Row row = result.one();
+      if (row != null) {
+        return row.getList(0, String.class).contains(keyspaceName);
       }
     }
     return false;
--- a/management-api-server/src/test/java/com/datastax/mgmtapi/NonDestructiveOpsIT.java
+++ b/management-api-server/src/test/java/com/datastax/mgmtapi/NonDestructiveOpsIT.java
@@ -32,6 +32,7 @@ import com.datastax.mgmtapi.resources.models.ReplicationSetting;
 import com.datastax.mgmtapi.resources.models.ScrubRequest;
 import com.datastax.mgmtapi.resources.models.Table;
 import com.datastax.mgmtapi.resources.models.TakeSnapshotRequest;
+import com.datastax.mgmtapi.resources.v2.models.TokenRangeToEndpointResponse;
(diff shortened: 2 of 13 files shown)


management-api-server/src/main/java/com/datastax/mgmtapi/resources/common/BaseResources.java
```diff
@@ -71,4 +73,21 @@ protected Response handle(Callable<Response> action) {
...
+      try {
+        ResultSet result =
+            app.cqlService.executePreparedStatement(
+                app.dbUnixSocketFile, "CALL NodeOps.getKeyspaces()");
+        Row row = result.one();
+        if (row != null) {
+          return row.getList(0, String.class).contains(keyspaceName);
+        }
+      } catch (Exception e) {
+        // cxonnection issue or some other problem. Eat it and return false.
```
Miles-Garnsey, line 88: Suggestion: Can we throw an exception here so that we can ultimately return a 500 error? I think false is a bit different to a genuine error, even though in this use case the difference is a bit subtle. One is (as you currently have) a 404, the other is a 500.

management-api-server/src/test/java/com/datastax/mgmtapi/NonDestructiveOpsIT.java
```diff
@@ -1045,12 +1039,35 @@ public void testMoveNode() throws IOException, URISyntaxException {
             });
   }
 
+  @Test
+  public void testGetTokenRangeToEndpointMap() throws Exception {
```
Miles-Garnsey, line 1043: Issue: Because this is a v2 endpoint, could the tests for it please sit alongside the tests for my repair endpoint (in test/java/com/datastx/mgmtapi/resources/v2`?

management-api-server/src/test/java/com/datastax/mgmtapi/NonDestructiveOpsIT.java
```diff
@@ -1080,4 +1097,14 @@ private Pair<Integer, String> responseAsCodeAndBody(FullHttpResponse r) {
...
   }
+
+  private int getNumTokenRanges() {
+    if (this.version.startsWith("3")) {
+      return 256;
+    }
+    if (this.version.startsWith("dse-68")) {
+      return 1;
+    }
+    return 16;
```
Miles-Garnsey, line 1108: Suggestion: it might be nice to have this explicitly check for Cassandra 4.x so that it is clear what we're trying to support.

Miles-Garnsey (changes requested):
This looks right to me, I need to manually test it before being 100% sure. I will do so tomorrow. There are 1-2 minor things in there that I've suggested/requested.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/common/BaseResources.java
emerkle826: We can. I initially just wanted this as a utility method that only returned `true` if the keyspace exists, and false otherwise. But we could leverage this in a future V2 `getKeyspace` endpoint. Right now, the `listKeyspaces` endpoint simply returns an empty list if the keyspace specified does not exist.

management-api-server/src/test/java/com/datastax/mgmtapi/NonDestructiveOpsIT.java
emerkle826: For unit tests (which I didn't write), that would be fine. This is an integration test though,

The integration tests have a bit of complexity with them as they are categorized as either "Isolated" or "Parallel". "Isolated" tests are ones that extend [BaseDockerIsolatedIntegrationTest](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-server/src/test/java/com/datastax/mgmtapi/BaseDockerIsolatedIntegrationTest.java) and do so because they create a non-predictable Cassandra state after they run and the state is difficult to restore. For example the "DestructiveOpsIT" tests and the "LifecycleIT" tests fall into this category. The main difference between these tests is that the Isolated ones will stop the Docker container running Cassandra after each test and start a new one for the next test so that results are predictable. The non-isolated tests spin up a Docker container and run all tests against the same container without stopping it in between tests.

If that were all, I could simply create an IT test in the new `v2` test package and be done. Except that isn't all. The IT tests are now run in [parallel]()https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/.github/workflows/ci.yaml#L14 in GitHub CI to try to speed things up. So if we create a new IT test in the `v2` package, I'll have to add it there as well, but for all the tests matrices (3.11, 4.0, 4.1, DSE, etc). By putting this test specifically in the `NonDestructiveOpsIT` test class, it will already run in CI.

So a follow-up question to you:
Do you want to add this new IT test class in the `v2` package and add it to the CI test matrices?

emerkle826: Actually, I might be able to do both. The way CI uses the test matrix is by specifying the test class name. So as long as I name the `v2` version of the Integration tests the same as one of the existing ones, I may not have to mess with the test matrix. Will test this locally and report back.

Miles-Garnsey: >  Do you want to add this new IT test class in the v2 package and add it to the CI test matrices?

Uh, sorry, I didn't quite understand what was going on here with it being an e2e test against an actual running management API. I don't mind too much now I know that. e2e tests don't strictly need to follow the structure/packaging of the production code in our other projects.

Having said that, well, it is nice to have smaller test classes I guess, so maybe this change is for the good. 

Suggestion/nit: Do you think we could call it `NonDestructiveOpsResourcesV2IT.java` though, obviously the naming of these tests (`IT` instead of `E2E`) is already confusing me a bit.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/common/BaseResources.java
Miles-Garnsey: I don't think this suggestion has been implemented, is it something we want to go for?

Miles-Garnsey (approved):
Approved.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/common/BaseResources.java
Miles-Garnsey: Ignore me, it has been implemented.

Miles-Garnsey:
Confirming that I've manually tested this against a real cluster and it seems to behave as expected:

```
$ curl 127.0.0.1:8080/api/v2/tokens/rangetoendpoint | jq
{
  "token_range_to_endpoints": [
    {
      "tokens": [
        -1070326447217483000,
        574332704784027300
      ],
      "endpoints": [
        "127.0.0.1"
      ]
    },
    {
      "tokens": [
        -3772959478813367000,
        -1070326447217483000
      ],
      "endpoints": [
        "127.0.0.2"
      ]
    },
    {
      "tokens": [
        -5592763185766151000,
        -3772959478813367000
      ],
      "endpoints": [
        "127.0.0.1"
      ]
    },
    {
      "tokens": [
        -8583309438055266000,
        -5592763185766151000
      ],
      "endpoints": [
        "127.0.0.2"
      ]
    },
    {
      "tokens": [
        3465377086168102400,
        5812374237656065000
      ],
      "endpoints": [
        "127.0.0.2"
      ]
    },
    {
      "tokens": [
        574332704784027300,
        3465377086168102400
      ],
      "endpoints": [
        "127.0.0.1"
      ]
    },
    {
      "tokens": [
        5812374237656065000,
        7352722417716966000
      ],
      "endpoints": [
        "127.0.0.2"
      ]
    },
    {
      "tokens": [
        7352722417716966000,
        -8583309438055266000
      ],
      "endpoints": [
        "127.0.0.1"
      ]
    }
  ]
}
```

The only nit left to address is the name of the test class, I've approved so it can be merged as soon as that is updated.

Miles-Garnsey:
@emerkle826 do you want to make that final change to the name of the test class and get this merged?

PR #385, opened 2023-09-20
Fix for Null statusChanges field in Job status

Miles-Garnsey:
From Reaper, we sometimes observe that the statusChanges field implemented as part of [this](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/338/files#diff-b6074c0b3803581492014197f0d6bff6c76c112a6c50056d20d532c0864108ce) PR can be null. This results in no repair progress notifications being passed to Reaper.

This PR:

1. Creates an integration test to detect the problem.
2. Attempts to fix the problem.


Miles-Garnsey:
On hold pending the ability to test the client library from inside this repo. https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/389

PR #388, opened 2023-09-20
Fix getJobStatus to correctly serialize the state changes

burmanm:


Miles-Garnsey (approved):
This PR is lacking any tests. However I have started writing one over [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/385) and it is only merging to the integration branch, so I'll approve and manually test from Reaper once merged.

PR #390, opened 2023-09-21
Test PR to see if removing enum from lists helps serialise Job type

Miles-Garnsey:
This PR is a test to see if the agent will build and run with a Job type returned from `getJobStatus()` which does not contain Lists of types containing enums. 

If it fails the next step will be to remove the `ProgressEventType` enum entirely and replace it with a string in all locations.


Miles-Garnsey:
Closing, we ended up hacking around this in #388 .

PR #393, opened 2023-09-26
Job Status Serialisation issue

Miles-Garnsey:
The Job constructor currently has a string type for the statusChanges parameter in the constructor and it appears that this causes RestEasy to output the statusChanges as a String in HTTP responses, instead of correctly serialising it into an Array<StatusChange> (as it is declared to be in the actual class body for Job). Thanks RestEasy.

Meanwhile, the annotations picked up for the OpenAPI document (which are subsequently propagated to the automatically generated client) pick up the statusChanges field (correctly) as a List<StatusChange> , which causes a deserialisation error on the client side when the client attempts to deserialise the JSON string primitive as an array. 

This PR harmonises the two by ensuring that any inputs to the `Job` constructor are declared as a real `List<StatusChange>` which necessitates some additional manipulation to deserialise the json from the RPC layer before re-serialising the whole thing as the correct JSON in the Resources layer.

This whole situation really needs a re-design.


burmanm:
Seems to work:

```
cassandra@cluster2-dc2-r1-sts-0 /]$ curl 'http://localhost:8080/api/v0/ops/executor/job?job_id=repair-0'
{"id":"repair-0","type":"repair","status":"COMPLETED","submit_time":1695758304316,"end_time":1695758304320,"error":null,"status_changes":[{"status":"COMPLETE","change_time":1695758304320,"message":""}]}
[cassandra@cluster2-dc2-r1-sts-0 /]$
```

burmanm (approved):

PR #402, opened 2023-10-11
Add async versions of flush and garbagecollect.

### LINKED ISSUE #403: Add async versions of flush and garbagecollect
Without these, the operations are quite unreliable and have to rely on timeouts. But garbagecollect could take a while and so on relying on cass-operator to just have infinite timeout makes finding errors difficult. And it also ties the operation cass-operator's lifecycle.

This is required by https://github.com/k8ssandra/cass-operator/issues/578 and https://github.com/k8ssandra/cass-operator/issues/580

burmanm:
Part of https://github.com/k8ssandra/cass-operator/issues/578 and https://github.com/k8ssandra/cass-operator/issues/580

Fixes #403

### DIFF
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/resources/v1/TableOpsResources.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/resources/v1/TableOpsResources.java
@@ -321,18 +321,18 @@ public class TableOpsResources extends BaseResources {
       responseCode = "202",
       description = "Job ID for table garbage collection process",
       content =
-      @Content(
-          mediaType = MediaType.TEXT_PLAIN,
-          schema = @Schema(implementation = String.class),
-          examples = @ExampleObject(value = "d69d1d95-9348-4460-95d2-ae342870fade")))
+          @Content(
+              mediaType = MediaType.TEXT_PLAIN,
+              schema = @Schema(implementation = String.class),
+              examples = @ExampleObject(value = "d69d1d95-9348-4460-95d2-ae342870fade")))
   @ApiResponse(
       responseCode = "400",
       description = "Invalid table garbage collection request",
       content =
-      @Content(
-          mediaType = MediaType.TEXT_PLAIN,
-          schema = @Schema(implementation = String.class),
-          examples = @ExampleObject(value = "tombstoneOption must be either ROW or CELL")))
+          @Content(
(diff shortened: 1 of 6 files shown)


management-api-server/src/main/java/com/datastax/mgmtapi/resources/v1/TableOpsResources.java
```diff
@@ -312,4 +312,105 @@ public Response list(
...
+        () -> {
+          List<String> tables = keyspaceRequest.tables;
+          if (CollectionUtils.isEmpty(tables)) {
+            tables = new ArrayList<>();
+          }
+
+          String keyspaceName = keyspaceRequest.keyspaceName;
+          if (StringUtils.isBlank(keyspaceName)) {
+            keyspaceName = "ALL";
+          }
```
emerkle826, line 352: issue:
If the request specifies a keyspace that does not exist, the Cassandra Storage service will throw an exception, which the NodeOpsProvider in the Agent will catch and throw a RuntimeException, which will bubble up in this `handle` call and result in one of our favorite `500 Internal Server Errors`. 

Could you add a check here and maybe return either a 404 or a 400 if the keyspace in the request does not exist? You can use the common method here to test if it exists:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-server/src/main/java/com/datastax/mgmtapi/resources/common/BaseResources.java#L85-L96
An example of how we use it is here:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-server/src/main/java/com/datastax/mgmtapi/resources/v2/TokenResourcesV2.java#L60-L62
Note that you will have to update the annotations if you send back a 404. I'm fine with a 400 in this case, so long as the message indicate the keyspace doesn't exist.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/v1/TableOpsResources.java
```diff
@@ -312,4 +312,105 @@ public Response list(
...
```
emerkle826, line 403: issue:
Same thing here with a non-existent keyspace.

emerkle826 (changes requested):
I would prefer to get 4XX responses if the requested keyspace does not exist.

emerkle826:
Closing this in favor of #409

PR #411, opened 2023-10-16
Serialize long types as Strings to avoid float64 conversion in Job

### LINKED ISSUE #412: 0.1.69 changed the serialization of JobDetails
We started to serialize long numbers as float64 in the JobDetails instead of string. This means the timestamps are suddenly floats and are not compatible with existing implementations such as cass-operator.

burmanm:
Fixes #412


Miles-Garnsey (approved):
Michael's sent me manual test results which show that the types are now set back to strings again. Approving.

emerkle826 (approved):

PR #414, opened 2023-10-25
Use tini for UBI8 images to ensure correct termination

### LINKED ISSUE #413: UBI8 images do not terminate correctly
In testing the UBI8 Cassandra images in cass-operator, it was observed that these images don't terminate the same way as the Ubuntu based images. This is likely due to how the image creates the process.

```[tasklist]
### DoD
[ ] - UBI8 based images should create the same style process as Ubuntu based images so that termination is the same
```

emerkle826:
Fixes #413


burmanm (approved):
"Works on my machine"

PR #423, opened 2023-11-16
Shade snakeyaml library

### LINKED ISSUE #422: Metrics endpoint fails to start for some DSE images
Some newer DSE images fail to startup the metrics endpoint due to library upgrades:
```
ERROR [DSE main thread] 2023-11-14 13:13:12,810  MetricsInterceptor.java:92 - Unable to start metrics endpoint
java.lang.NoSuchMethodError: org.yaml.snakeyaml.parser.ParserImpl.<init>(Lorg/yaml/snakeyaml/reader/StreamReader;)V
	at com.fasterxml.jackson.dataformat.yaml.YAMLParser.<init>(YAMLParser.java:178)
	at com.fasterxml.jackson.dataformat.yaml.YAMLFactory._createParser(YAMLFactory.java:471)
	at com.fasterxml.jackson.dataformat.yaml.YAMLFactory.createParser(YAMLFactory.java:375)
	at com.fasterxml.jackson.dataformat.yaml.YAMLFactory.createParser(YAMLFactory.java:15)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3642)
	at io.k8ssandra.metrics.config.ConfigReader.readConfig(ConfigReader.java:38)
	at io.k8ssandra.metrics.interceptors.MetricsInterceptor.intercept(MetricsInterceptor.java:68)
	at org.apache.cassandra.service.CassandraDaemon.start(CassandraDaemon.java)
	at com.datastax.bdp.server.DseDaemon.start(DseDaemon.java:559)
	at org.apache.cassandra.service.CassandraDaemon.activate0(CassandraDaemon.java:830)
	at org.apache.cassandra.service.CassandraDaemon.access$100(CassandraDaemon.java:93)
	at org.apache.cassandra.service.CassandraDaemon$3.run(CassandraDaemon.java:738)
```

### LINKED ISSUE #425: Shade Jackson libraries in agents
In light of #422, it would be safer to shade all of the Jackson libraries we use in the Management API agents so that we do not have to rely on the versions shipped with Cassandra/DSE. Shading the snakeyaml library will fix things related to #422, but shading the Jackson libraries used will hopefully avoid future library conflicts.

emerkle826:
Fixes #422
Fixes #425


management-api-shaded-dataformat-yaml/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+  <artifactId>datastax-mgmtapi-shaded-dataformat-yaml</artifactId>
+  <profiles>
+    <profile>
+      <id>default</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>com.fasterxml.jackson.dataformat</groupId>
```
burmanm, line 26: I'm not sure if this is enough to shade only. While we only use dataformat directly, we pull other Jackson dependencies from Cassandra.

For example, mvn dependency:tree for 5.0 agent gives us:

```
[INFO] io.k8ssandra:datastax-mgmtapi-agent-5.0.x:jar:0.1.0-SNAPSHOT
[INFO] +- io.k8ssandra:datastax-mgmtapi-common:jar:0.1.0-SNAPSHOT:compile
<clip>
[INFO] +- io.k8ssandra:datastax-mgmtapi-agent-common:jar:0.1.0-SNAPSHOT:compile
<clip>
[INFO] |  \- io.k8ssandra:datastax-mgmtapi-shaded-dataformat-yaml:jar:0.1.0-SNAPSHOT:compile
[INFO] |     \- com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:jar:2.15.3:compile
<clip>
[INFO] \- org.apache.cassandra:cassandra-all:jar:5.0-alpha2:provided
<clip>
[INFO]    +- com.fasterxml.jackson.core:jackson-core:jar:2.13.2:compile
[INFO]    +- com.fasterxml.jackson.core:jackson-databind:jar:2.13.2.2:compile
[INFO]    +- com.fasterxml.jackson.core:jackson-annotations:jar:2.13.2:compile
[INFO]    +- com.fasterxml.jackson.datatype:jackson-datatype-jsr310:jar:2.13.2:provided
[INFO]    +- com.boundary:high-scale-lib:jar:1.0.6:provided
[INFO]    +- org.yaml:snakeyaml:jar:1.26:provided
```

Could we end up with some conflicts? I also noticed 5.x is using different dropwizard-metrics, so that might be the next thing that breaks us..

emerkle826: > While we only use dataformat directly, we pull other Jackson dependencies from Cassandra.
Good point. Technically, I'm only wrapping Jackson dataformat-yaml, and only shading snakeyaml. To be safe, we probably need to shade dataformat-yaml, and any of its depenedencies that may be provided by cassandra-all/dse-all. From the looks of it, that would include jackson-annotations, jackson-core and jackson-databind. 

> Could we end up with some conflicts?
Always a possibility with this project. The reason we have so many "agent" sub-modules is because of the changes to the internal Cassandra classes. But we also have the possibility of version incompatibility with the "provided" libraries as well. Until now, we have gotten away with it.

> I also noticed 5.x is using different dropwizard-metrics, so that might be the next thing that breaks us..
So if we get into issues with this, the next step is to break the metrics stuff out into the different agent sub-modules so it can be compiled against the versions of the libraries that will actually be provided by the version of Cassandra/DSE that the agent is for. Which means refactoring all the metrics stuff so that only the common parts stay in agent-common, and the runtime classes are instantiated as necessary via SPI. That is doable, but is not a fun task.

emerkle826: So I just created #425 to address the Jackson library stuff. Shading just snakeyaml for now should fix the current issue and buy us some time.

emerkle826: @burmanm I also created #426 to cover dropwizard shading

PR #428, opened 2023-11-20
Modify the bucket reading behavior, try to fetch the bucketOffsets fo…

### LINKED ISSUE #444: Latency metrics aren't exposed anymore in 4.1
Upgrading to Cassandra 4.1 will make some latency metrics disappear because the bucket offsets have changed in that version.
We need to address that issue in the metrics endpoint so that we get those metrics back in the output.

burmanm:
…r OSS Cassandra versions also and deal with the overflow buckets

Fixes #444

### DIFF
--- a/management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricRegistryListener.java
+++ b/management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricRegistryListener.java
@@ -22,6 +22,7 @@ import com.codahale.metrics.Snapshot;
 import com.codahale.metrics.Timer;
 import io.k8ssandra.metrics.config.Configuration;
 import io.prometheus.client.Collector;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricRegistryListener.java
```diff
@@ -373,20 +389,24 @@ private void setTimerFiller(
...
+          } else {
+            logger.debug(
+                String.format(
+                    "Unknown type for %s, wants: %s\n", proto.getMetricName(), snapshotClass));
           }
 
           // This can happen if histogram isn't EstimatedDecay or EstimatedHistogram
-          if (values.length != buckets.length) {
-            logger.debug(
+          if (values.length > buckets.length + 1 || values.length < buckets.length) {
```
emerkle826, line 399: Question:
with this logic, it means that `values.length` can equal `buckets.length` OR `buckets.length + 1`. Is that what we want/need here?

emerkle826 (approved):
Just the one question/clarification. As a general comment, I'd like to try to refactor things to get rid of the reflection, but last time I looked at doing that here, it was not going to be easy to pull off. I'll create a refactor ticket if we don't have one and maybe one day I'll try to tackle that. Otherwise this looks good.

adejanovski:
@burmanm, are we good for a merge here?

PR #447, opened 2024-02-05
Scrub is not functioning with keyspaceName "ALL"

burmanm:
Unlike the REST API suggests (and tests), the keyspaceName "ALL" is not accepted by the NodeOpsProvider. 

Fixes https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/448


emerkle826 (changes requested):
This looks fine. It would be nice to add an integration test, but it's an async job and I don't know if there is any easy way to tell that a scrub has been performed or not. We have a test that scrubs a single table in a single keyspace, and the Integration test there essentially just ensures that the call to scrub succeeds.

[issue]
The only issue I ask is to add an update to the [Changelog](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/CHANGELOG.md). It helps when I cut out the releases as I can just copy that into the release description.

PR #450, opened 2024-02-06
TableOpsResources should not use compactRequest.tables

### LINKED ISSUE #437: Compaction throws NullPointerException but returns 202 Accepted
In the context of cass-operator, it seems that invoking the `/api/v1/ops/tables/compact` endpoint returns a 202 Accepted, but cass-operator logs show that the task failed due to a NullPointerException:

`"error": "task failed: java.lang.NullPointerException"`

burmanm:
We already had a "tables" variable, which was empty ArrayList in case the compactRequest.tables was null. This just wasn't used if tokenRange wasn't set.

Also, add logging in case we get Throwable so it makes debugging easier.

Fixes #437 

Tested manually:

```
➜  cass-operator git:(master) ✗ kubectl get cassandratask                                     
NAME              DATACENTER   JOB       SCHEDULED   STARTED   COMPLETED
example-compact   dc2          compact               41s       31s
➜  cass-operator git:(master) ✗
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Changelog for Management API, new PRs should update the `main / unreleased` sect
 
 ## unreleased
 
+* [BUGFIX] [#448](https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/448) Scrub did not work when targetting all the keyspaces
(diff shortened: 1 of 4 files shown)


emerkle826 (changes requested):
This is fine, just need a Changelog entry

Issue #292, opened 2023-04-24
Time out on taking snapshot


shaotingcheng:
When calling the management API `/api/v0/ops/node/snapshots` to take the snapshot, I encounter the 30s timeout error
`Query timed out after PT30S`

```
curl -k -L -XPOST -H "Content-Type:application/json" 'http://localhost:8080/api/v0/ops/node/snapshots' -d '{"snapshot_name": "test124"}'
Query timed out after PT30S
```

From the container logs
```
INFO  [nioEventLoopGroup-2-1] 2023-04-24 04:32:38,254 Cli.java:617 - address=/127.0.0.1:44180 url=/api/v0/ops/node/snapshots status=500 Internal Server Error
INFO  [nioEventLoopGroup-2-1] 2023-04-24 04:32:38,270 Cli.java:617 - address=/10.100.42.232:42336 url=/api/v0/probes/liveness status=200 OK
ERROR [nioEventLoopGroup-3-3] 2023-04-24 04:32:38,336 NodeOpsResources.java:562 - Error when executing request
com.datastax.oss.driver.api.core.DriverTimeoutException: Query timed out after PT30S
	at com.datastax.oss.driver.api.core.DriverTimeoutException.copy(DriverTimeoutException.java:34)
	at com.datastax.oss.driver.internal.core.util.concurrent.CompletableFutures.getUninterruptibly(CompletableFutures.java:149)
	at com.datastax.oss.driver.internal.core.cql.CqlRequestSyncProcessor.process(CqlRequestSyncProcessor.java:53)
	at com.datastax.oss.driver.internal.core.cql.CqlRequestSyncProcessor.process(CqlRequestSyncProcessor.java:30)
	at com.datastax.oss.driver.internal.core.session.DefaultSession.execute(DefaultSession.java:230)
	at com.datastax.oss.driver.api.core.cql.SyncCqlSession.execute(SyncCqlSession.java:54)
	at com.datastax.oss.driver.api.core.cql.SyncCqlSession.execute(SyncCqlSession.java:78)
	at com.datastax.mgmtapi.CqlService.executeCql(CqlService.java:35)
	at com.datastax.mgmtapi.resources.K8OperatorResources.lambda$checkReadiness$0(K8OperatorResources.java:87)
	at com.datastax.mgmtapi.resources.NodeOpsResources.handle(NodeOpsResources.java:554)
[...]

PR #454, opened 2024-03-11
Use a longer driver timeout for drain

### LINKED ISSUE #453: Use a longer driver timeout for drain
In some scenarios, the drain operation can take longer than 30 seconds (the [configured driver requesttimetout](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-server/src/main/java/com/datastax/mgmtapi/UnixSocketCQLAccess.java#L131)). Ideally, the call to drain a node would be asynchronous and a new endpoint could be added to poll the status of the drain. However, it would maybe be beneficial to simply adding a "slow" driver execution profile for this, and change the implementation of drain to make a CQL query with the slow profile enabled (see the [Java driver docs](https://docs.datastax.com/en/developer/java-driver/4.17/manual/core/configuration/#execution-profiles) for more details on execution profiles).

This "slow" execution profile could also be used for other Management API functions that could take longer than 30 seconds to complete under the hood, with the benefit of being able to make the change to management API and not having to make changes on down-stream projects (like cass-operator).

emerkle826:
Fixes #453

### DIFF
--- a/management-api-server/src/main/resources/application.conf
+++ b/management-api-server/src/main/resources/application.conf
@@ -1,0 +1,7 @@
+datastax-java-driver {
+  profiles {
+    slow {
+      basic.request.timeout = 180 seconds
+    }
+  }
+}
--- a/management-api-server/src/main/resources/application.conf
+++ b/management-api-server/src/main/resources/application.conf
@@ -1,7 +1,10 @@
 datastax-java-driver {
   profiles {
+    ## The slow profile below sets the driver timeout to 0 seconds
+    ## to effectively disable the query timeout all together. This
+    ## is not something you would normally want to do with the driver.
     slow {
-      basic.request.timeout = 180 seconds
+      basic.request.timeout = 0 seconds
     }
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changelog for Management API, new PRs should update the `main / unreleased` sect
 ```
 
 ## unreleased
+* [FEATURE] [$453](https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/453) Use a longer driver timeout for drain
 
 ## v0.1.73 (2024-02-20)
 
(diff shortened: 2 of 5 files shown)


management-api-server/src/main/resources/application.conf
```diff
@@ -0,0 +1,7 @@
+datastax-java-driver {
+  profiles {
+    slow {
+      basic.request.timeout = 180 seconds
```
burmanm, line 4: Do we need to have this set as hardcoded value? At the moment, the way we call /drain might benefit from even no timeout at all, since the pod has a strict timeline until it gets killed in any case.

emerkle826: I'd have to see if it can be set with a variable or something at runtime. The idea is to make this large enough so that none of the driver "calls" times out before getting a response. I don''t think there is a way to set no timeout for driver requests, assuming you mean wait forever. If Kubernetes things end up calling this and they have their own timeout, then as long is this is at least as long as K8s will wait, we should be fine. I'm still not sure the full context of when `/drain` is called and waiting longer than 3 minutes is expected.

We can raise this timeout to 5 or 10 minutes if you think that would be better. It won't hurt any drain calls that finish in less time. Also, the drain call is the only one using this new "slow" query thing for now. Although #292 might also benefit from using the slow query for taking a snapshot if that function is also taking a long time.

emerkle826: Actually, it seems that we can configure the timeout to 0, which would effectively make it wait forever. Do we want to do this?

burmanm: Users are able to configure the /drain timeout using gracePeriod for pods.

emerkle826: This change is only going to affect the internal driver query for the drain operation. Outside of a kubernetes environment, with this set to 0, a call to /drain won't get a driver query request timeout. Something else may happen that causes the operation to fail, but otherwise, the driver will wait forever for the agent to complete and return.

burmanm (approved):

PR #466, opened 2024-04-02
Add tar to DSE images

### LINKED ISSUE #465: Add `tar` to the DSE images
When troubleshooting misbehaving pods it's often required to be able to extract a large number of files from the container outside of it.
`tar` is very handy to perform such operations and is present in the Cassandra images but not the DSE ones.
We'd need to add `tar` to the DSE images as well. It is also mandatory to be able to use `kubectl cp` commands.

```[tasklist]
### Definition of Done
- [ ] `tar` is added to the dse images
```

emerkle826:
Fixes #465

### DIFF
--- a/.github/workflows/add_to_project.yaml
+++ b/.github/workflows/add_to_project.yaml
@@ -10,7 +10,7 @@ jobs:
     name: Add issue to GH project
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/add-to-project@v0.5.0
+      - uses: actions/add-to-project@v1
         with:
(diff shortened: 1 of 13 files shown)


emerkle826:
The `tar` binary is part of the Ubuntu based images, it was just missing in the UBI based image

adejanovski (approved):
Cool stuff, thanks!

emerkle826:
The verification/tests aren't reliable. Right now it tries to test for the absence or existence of a file. And then it checks the Docker exec process return code. Sometimes that return code is null. I need a better way to verify this stuff.

PR #478, opened 2024-04-26
Fix cassandra-topology.properties removal

### LINKED ISSUE #477: Fix `cassandra-topology.properties` removal
Commit https://github.com/k8ssandra/management-api-for-apache-cassandra/commit/1c79885cec838bfedc73b802f7b63ec6f3c7de0c attempted to remove the topology file from the images by doing so in the entrypoint script. This seems to fail due to ownership/permissions issues in some environments (at least in K8s with cass-operator).

We should put the file removal in the Dockerfiles themselves instead of the entrypoint script.

emerkle826:
Fixes #477

### DIFF
--- a/management-api-server/src/test/java/com/datastax/mgmtapi/BaseDockerIntegrationTest.java
+++ b/management-api-server/src/test/java/com/datastax/mgmtapi/BaseDockerIntegrationTest.java
@@ -128,7 +128,7 @@ public abstract class BaseDockerIntegrationTest {
     if (docker != null) {
       temporaryFolder.delete();
       temporaryFolder.create();
-      docker.startManagementAPI(version, getEnvironmentVars());
+      docker.startManagementAPI(version, getEnvironmentVars(), getUser());
     }
   }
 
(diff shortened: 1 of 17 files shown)


emerkle826:
@burmanm I have tested images created with this PR in the cass-operator tests and Management API starts up correctly.

management-api-server/src/test/java/com/datastax/mgmtapi/BaseDockerIntegrationTest.java
```diff
@@ -167,6 +167,15 @@ protected ArrayList<String> getEnvironmentVars() {
         "DSE_MGMT_EXPLICIT_START=true");
   }
 
+  protected String getUser() {
+    // The default should be either "cassandra:root" or "dse:root"
+    // Subclasses can override this to test alternate user behavior
+    if (this.version.startsWith("dse")) {
```
burmanm, line 173: "dse:root" ?

The DSE test fails now.

emerkle826: The DSE tests are passing in this PR now. I also loaded the image into my Kind cluster for cass-operator and ran the tests against that and it is working for me.

burmanm (approved):

PR #485, opened 2024-05-15
Rename DSE7 to HCD

### LINKED ISSUE #484: Rename DSE7 to HCD
DSE7 is rebranded to HCD. We should update the DSE7 artifacts to be labeled HCD.

emerkle826:
Fixes #484


management-api-agent-hcd/README.md
```diff
@@ -0,0 +1,56 @@
+# Management API with HCD (Converged Core)
```
tiagomlalves, line 1: Could we have here instead:
```
# Management API with HCD (Hyper-Converged Database)
```

emerkle826: Absoultely

PR #491, opened 2024-05-31
Allow readOnlyRootFilesystem support in UBI8 images

### LINKED ISSUE #521: Install mgmt-api agent in the Dockerfile, not docker-entrypoint.sh
The cassandra-env.sh should not be modified in the entrypoint.sh if we want to run readOnlyRootFilesystem.

burmanm:
@emerkle826 

Fixes #521


emerkle826 (approved):
This looks fine since for 4.1.x and 5.0.x UBI images, nothing else on the root file system is going to be modified. If we have to also support 4.0.x, a bit more from the entrypoint script will have to be brought over

emerkle826 (commented):
@burmanm just needs a Changelog entry ;)

PR #508, opened 2024-07-01
Use Set as the CassandraMetricDefinition sample holder

burmanm:
Fixes https://github.com/k8ssandra/k8ssandra-operator/issues/1345

### DIFF
--- a/management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricDefinition.java
+++ b/management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricDefinition.java
@@ -7,14 +7,13 @@ package io.k8ssandra.metrics.builder;
 
 import io.prometheus.client.Collector;
 import java.util.List;
+import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
-import org.slf4j.LoggerFactory;
 
 public class CassandraMetricDefinition
-    implements Consumer<List<Collector.MetricFamilySamples.Sample>> {
-  private static final org.slf4j.Logger logger =
-      LoggerFactory.getLogger(CassandraMetricDefinition.class);
+    implements Consumer<List<Collector.MetricFamilySamples.Sample>>,
+        Comparable<CassandraMetricDefinition> {
 
   private final List<String> labelNames;
   private final List<String> labelValues;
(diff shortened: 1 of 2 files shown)


management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricDefinition.java
```diff
@@ -74,4 +73,46 @@ public void setKeep(boolean keep) {
...
+        + ", filler="
+        + filler
+        + '}';
+  }
+
+  @Override
+  public int compareTo(CassandraMetricDefinition o) {
+    return metricName.compareTo(o.metricName)
+        + (labelNames.hashCode() - o.labelNames.hashCode())
+        + (labelValues.hashCode() - o.labelValues.hashCode());
```
emerkle826, line 116: Issue:
The hashcode math here could make the behavior of `compareTo` unpredictable. It's possible that the `metricName` compareTo returns a value that is orders of magnitude smaller than the compareTo value returned from the `labelNames` and `labelValues` lists. Isolated, using the difference between the hashcodes of either List should work well enough to order things alphabetically. But when you start adding the hascode values together, it will shift things. The end result will be that, on a given comparison, the list that has the largest change will drive the aggregate result, but that won't make the ordering predictable.

Maybe, in practice and in the context of the metrics, the result is more predictable than just random lists of strings and random metric names, but it seems like this could become a problem at some point.

emerkle826 (changes requested):

management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricDefinition.java
burmanm: Well, it doesn't really matter what the order is in this case - as long as it can be inserted to a Set. Random list is just fine as long as the Set stays in correct format.

But, I made the hashCode now follow what java.util.AbstractList does as this is a List itself in reality (metricName + labelNames + labelValues).

emerkle826 (approved):

emerkle826 (changes requested):
It looks like some of the tests are failing. I've tried re-running them, but they are still failing.

burmanm:
Well, now.. I modified the wrong method.

emerkle826 (approved):
Ok, now this looks good. Good catch!

PR #509, opened 2024-07-02
Fix the SSLContext to point to the right context 

### LINKED ISSUE #507: SSL Context Hot Reloading doesn't work as expected
[SSLContext reload](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/376) does not seem to be updated when cert files are changed. It seems that even though a log message is displayed that changes have been detected in the files, the actual SSLContext for the server remains unchanged.

Steps to reproduce.
1. Create a short lived cert that will expire in a few minutes to use as a cert for the server
2. Update cert to a longer lived cert
3. Verify the `Detected change in the SSL/TLS certificates, reloading` log occurs
4. Execute a local curl command e.g. `curl --cert tls.crt --key tls.key --cacert ca.crt https://localhost:8080/api/v0/lifecycle/pid` with updated certs
5. Output of above command will be `SSL certificate problem: certificate has expired` even though the certs should have been updated to the long-lived cert

Restarting the management api process (and therefore restarting Cassandra) will result in the process picking up the new cert, however, it would be great if the restart was not necessary

burmanm:
SSLContext should be refreshed to point to the right context, Instead of the old one when hot reloading SSL.

Fixes #507

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,10 @@ Changelog for Management API, new PRs should update the `main / unreleased` sect
 
 ## unreleased
 
+* [BUGFIX] [#507]((https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/507) Fix the SSLContext reloading in the initChannel when certificates have changed.
+
 ## v0.1.82 (2024-06-14)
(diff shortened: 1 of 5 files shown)


emerkle826 (approved):

emerkle826:
@burmanm Can you add a CHANGELOG entry please.

PR #523, opened 2024-08-20
Trust store reload endpoint and node ops method

### LINKED ISSUE #483: tsreload endpoint
dsetool has a tsreload command which force-reloads the truststore to take account of new certificates which may have been added.

We should add a management API endpoint that can be called to execute this operation from maangement API. The bean is called `DseReloadableTrustManagerMXBean`.

Miles-Garnsey:
Fixes #483


burmanm:
As noted in the Slack, I'll add it here also: Add a new item to the FeatureSet to indicate this API endpoint is available.

management-api-agent-dse-6.8/pom.xml
```diff
@@ -45,9 +45,27 @@
     </dependency>
     <dependency>
       <groupId>com.datastax.dse</groupId>
-      <artifactId>dse-commons</artifactId>
+      <artifactId>dse-core</artifactId>
```
burmanm, line 48: @emerkle826 You might have more insight to review this dependency change. I don't see in this PR why this was done.

management-api-agent-dse-6.8/src/main/java/com/datastax/mgmtapi/shim/DseAPI68.java
```diff
@@ -333,4 +334,8 @@ public RpcStatementShim makeRpcStatement(String method, String[] params) {
   public HintsService getHintsService() {
     return HintsService.instance;
   }
+
+  public void reloadTrustManager() throws Exception {
+    DseReloadableTrustManager.serverEncryptionInstance().reloadTrustManager();
```
burmanm, line 339: question: Is there a separate clientEncryptionInstance() that might need reload also?

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
```diff
@@ -82,4 +82,8 @@ default Object handleRpcResult(Callable<Object> rpcResult) throws Exception {
   default List<String> getKeyspaces() {
     return StorageService.instance.getKeyspaces();
   }
+
+  default void reloadTrustManager() throws Exception {
+    throw new Exception("Unimplemented for Cassandra, only available for DSE");
```
burmanm, line 87: suggestion: UnsupportedOperationException instead?

burmanm: That is also RuntimeException (since this is a Runtime issue) so the method does not have to declare throw Exception.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
```diff
@@ -663,6 +663,37 @@ public Response searchIndexRebuild(
         });
   }
 
+  @POST
+  @Path("/reload-truststore")
```
burmanm, line 667: discussion/suggestion/question/<I'm not sure if there's a correct type annotation>: The use of "-" here is a bit weird compared to the rest of the methods in mgmt-api. For similar reloads we already "schema/reload", "seeds/reload" vs. "reload-truststore". Could we instead have "/truststore/reload" ?

management-api-agent-dse-6.8/pom.xml
emerkle826: @burmanm This dependency change was made to allow for access to class `com.datastax.bdp.transport.common.DseReloadableTrustManager`. `dse-commons` provides 4 classes in package `com.datastax.bdp.transport.common`, but not `DseReloadableTrustManager`. I need to double check to make sure that the classes in `dse-commons` are also in `dse-core`.

emerkle826: Also, since these dependencies are listed as provided, we should be okay here as we would have bigger issues if `dse-commons` or `dse-core` are not available within the JVM running DSE.

management-api-agent-dse-6.8/src/main/java/com/datastax/mgmtapi/shim/DseAPI68.java
emerkle826: There is, but the implementation seems to be the same as `serverEncryptionInstance()`. @Miles-Garnsey it might be good to double check me on that.

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
emerkle826: I agree with this.

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
```diff
@@ -82,4 +83,11 @@ default Object handleRpcResult(Callable<Object> rpcResult) throws Exception {
   default List<String> getKeyspaces() {
     return StorageService.instance.getKeyspaces();
   }
+
+  default void reloadInternodeEncryptionTruststore() throws Exception {
+    throw new Exception("Unimplemented for Cassandra, only available for DSE");
+  };
+  default X509Certificate[] getEncryptionTruststoreIssuers() throws Exception {
```
emerkle826, line 90: suggestion:
Shoudl probably also change this to `UnsupportedOperationException` as well

emerkle826 (changes requested):

management-api-agent-dse-6.8/pom.xml
emerkle826: issue:
@Miles-Garnsey I double checked and there are classes in `dse-commons` that are _not_ in `dse-core`, though both artifacts provide classes in some of the same packages. That is something that should probably be addressed in the DSE code base, but that outside the scope here.

We should probably keep the `dse-commons` dependency and just add `dse-core` as you have it here.

emerkle826 (changes requested):

management-api-server/doc/openapi.json
```diff
@@ -872,6 +872,25 @@
         "summary" : "Move node on the token ring to a new token. This operation returns immediately with a job id."
       }
     },
+    "/api/v0/ops/node/reload-truststore" : {
```
burmanm, line 875: issue: This is still different than ``@PATH``

emerkle826: This should be auto-generated during the compile phase, so long as you don't specify `-DskipOpenApi`. Either the skip option was used, or you forgot to add this file to your most recent commit

Miles-Garnsey:
> @Miles-Garnsey I double checked and there are classes in dse-commons that are not in dse-core, though both artifacts provide classes in some of the same packages. That is something that should probably be addressed in the DSE code base, but that outside the scope here.

@emerkle826 is there some reason why we need both core and commons? It doesn't seem that anything from core is used which isn't in commons, and I don't know how they'll play together given both artifacts contain the same class names....

management-api-common/src/main/java/com/datastax/mgmtapi/shims/CassandraAPI.java
Miles-Garnsey: I'll see what I can do.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
Miles-Garnsey: I've made some changes to allow for the fact that we may one day have an endpoint to do things with client encryption too. As a result I've also adopted something similar to this.

management-api-agent-dse-6.8/pom.xml
Miles-Garnsey: @emerkle826 is there some reason why we need both core and commons? It doesn't seem that anything from core is used which isn't in commons, and I don't know how they'll play together given both artifacts contain the same class names....

management-api-agent-dse-6.8/src/main/java/com/datastax/mgmtapi/shim/DseAPI68.java
Miles-Garnsey: I don't want to get into this. It has nothing to do with getting CA rotation working and should be discussed elsewhere.

management-api-agent-dse-6.8/pom.xml
emerkle826: @Miles-Garnsey we don't need both. I've tested it a little, and since all of these dependencies are marked as `provided`, we would only need them if we are referencing classes defined within them. We do need `dse-db`, and your chnages here require `dse-core`, so you don't have to bring back `dse-common`

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -940,6 +941,22 @@ public String move(
...
+  }
+
+  @Rpc(name = "getInternodeEncryptionTruststoreIssuers")
+  public List<String> getEncryptionTruststoreIssuers() throws JsonProcessingException, Exception {
+    List<X509Certificate> issuers =
+        Arrays.stream(ShimLoader.instance.get().getEncryptionTruststoreIssuers())
+            .collect(Collectors.toList());
+    ObjectMapper objectMapper = new ObjectMapper();
+    return issuers.stream()
+        .map(i -> objectMapper.writeValueAsString(i))
```
emerkle826, line 956: issue:
`writeValueAsString` throws checked exceptions, but you can't call methods that throw checked exceptions within a lambda expression, so this will need to be re-written

emerkle826 (changes requested):

management-api-server/doc/openapi.json
Miles-Garnsey: Regenerated.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
Miles-Garnsey: Thanks for the feedback, this is now fixed.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
```diff
@@ -663,6 +663,75 @@ public Response searchIndexRebuild(
...
+            // rebuilding search index is only possible on DSE
+            return Response.status(Response.Status.BAD_REQUEST)
+                .entity("Reloading the truststore manually is only possible on DSE")
+                .build();
+          }
+
+          app.cqlService.executeCql(
+              app.dbUnixSocketFile, "CALL NodeOps.reloadInternodeEncryptionTruststore()");
+
+          return Response.ok("OK").build();
```
emerkle826, line 694: suggestion:
If something goes wrong reloading the truststore, the exception is going to be masked in the API call a bit and just return a 500 Internal Server Error, and it will require digging into the Cassandra/DSE logs to figure out what actually happened. It would be better to catch things here and log them before returning a 500.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
```diff
@@ -940,6 +941,29 @@ public String move(
     return submitJob("move", moveOperation, async);
   }
 
+  @Rpc(name = "reloadInternodeEncryptionTruststore")
+  public void reloadInternodeEncryptionTruststore() throws Exception {
+    ShimLoader.instance.get().reloadInternodeEncryptionTruststore();
+  }
+
+  @Rpc(name = "getInternodeEncryptionTruststoreIssuers")
+  public List<String> getEncryptionTruststoreIssuers() throws Exception {
```
burmanm, line 950: question Why is this feature required for CA rotation?

Miles-Garnsey: It will be required for e2e testing this inside this repo. We may not do that just yet (depending on time constraints) but if we don't include it there is no way to introspect the contents of the truststore as Cassandra sees them.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
```diff
@@ -663,6 +663,75 @@ public Response searchIndexRebuild(
...
+  @Operation(summary = "get internode trustsstore", operationId = "getInternodeTrustStoreIssuers")
+  public Response getInternodeTruststoreIssuers() {
+    return handle(
+        () -> {
+          final String releaseVersion =
+              ResponseTools.getSingleRowStringResponse(
+                  app.dbUnixSocketFile, app.cqlService, CASSANDRA_VERSION_CQL_STRING);
+          if (!releaseVersion.startsWith("4.0.0.68") && !releaseVersion.startsWith("4.0.0.69")) {
+            // rebuilding search index is only possible on DSE
+            return Response.status(Response.Status.BAD_REQUEST)
```
emerkle826, line 718: issue:
again, the possible 400 returned here isn't documented in the annotation above.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
```diff
@@ -663,6 +663,75 @@ public Response searchIndexRebuild(
...
+  @GET
+  @Path("/encryption/internode/truststore/issuers")
+  @Produces(MediaType.TEXT_PLAIN)
+  @ApiResponse(
+      responseCode = "200",
+      description = "Got list of truststore issuers",
+      content =
+          @Content(
+              mediaType = MediaType.TEXT_PLAIN,
+              schema = @Schema(implementation = String.class),
```
emerkle826, line 707: issue:
If we really need the entire X509Certificate object represented as JSON to be returned from the API, then a model should be created in `com.datastax.mgmtapi.resources.models`, similar to how we handle things like [SnapshotDetails](com.datastax.mgmtapi.resources.models). Then your `@Produces` line above would specify [APPLICATION_JSON](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java#L353) and the `schema` line here would specify that new model class, not `String.class`.

I know that is a lot more work, but that is the goal for many of the API endpoints in here. Existing endpoints get a bit of a pass since there are existing consumers we don't want to break compatibility with. But for new endpoints that return something less trivial, we should create the proper model classes and schemas here for the API.

emerkle826 (changes requested):
@Miles-Garnsey @burmanm After discussing with Miles, I don't think exposing the server's trust store is something normal API consumers need. It poses a security risk and the intended use was for testing/verification. Let's remove all the bits that return truststore info and trim this down to just a call to rotating the certificates.

emerkle826:
@Miles-Garnsey please add a [Changelog](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/CHANGELOG.md) entry as well when you get this all sorted.

management-api-agent-common/src/main/java/com/datastax/mgmtapi/NodeOpsProvider.java
burmanm: This repo is not meant to test DSE features though, if you use the same API as nodetool, then we can assume it just works. 

Also, since we didn't want to include client truststore information, we shouldn't add anything else that isn't used right now and not related to CA rotation.

There's other issue with it exposing also the truststore content outside the security realm of how Secrets are handled.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
Miles-Garnsey: How can I catch an exception from the RPC `NodeOps.reloadInternodeEncryptionTruststore()`? I don't think exceptions can be passed through the boundary of the RPC call? 

Are you proposing that we return some string from `NodeOps.reloadInternodeEncryptionTruststore()` stating whether the operation errored?

emerkle826: Looking at the other endpoint implementations, it looks like many do not document the generic 500 error in these cases, so let's just leave it as is.

management-api-agent-dse-6.8/src/main/java/com/datastax/mgmtapi/shim/DseAPI68.java
```diff
@@ -20,13 +21,8 @@
...
 import java.net.InetAddress;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.security.cert.X509Certificate;
```
emerkle826, line 24: issue:
remove unused import

management-api-agent-dse-6.9/src/main/java/com/datastax/mgmtapi/shim/DseAPI69.java
```diff
@@ -20,6 +21,7 @@
 import java.lang.reflect.Field;
 import java.net.InetAddress;
 import java.nio.charset.StandardCharsets;
+import java.security.cert.X509Certificate;
```
emerkle826, line 24: issue:
remove unused import

management-api-server/src/main/java/com/datastax/mgmtapi/resources/NodeOpsResources.java
```diff
@@ -663,6 +663,38 @@ public Response searchIndexRebuild(
...
+  @Path("/encryption/internode/truststore/reload")
+  @Produces(MediaType.TEXT_PLAIN)
+  @ApiResponse(
+      responseCode = "200",
+      description = "Truststore reloaded successfully",
+      content =
+          @Content(
+              mediaType = MediaType.TEXT_PLAIN,
+              schema = @Schema(implementation = String.class),
+              examples = @ExampleObject(value = "OK")))
```
emerkle826, line 676: :suggestion
Add the 400 response annotation for unsupported server versions
```suggestion
  @ApiResponse(
      responseCode = "200",
      description = "Truststore reloaded successfully",
      content =
          @Content(
              mediaType = MediaType.TEXT_PLAIN,
              schema = @Schema(implementation = String.class),
              examples = @ExampleObject(value = "OK")))
  @ApiResponse(
      responseCode = "400",
      description = "Unsupported Operation",
      content =
          @Content(
              mediaType = MediaType.TEXT_PLAIN,
              schema = @Schema(implementation = String.class),
              examples =
                  @ExampleObject(
                      value =
                          "Reloading the truststore manually is only possible on DSE")))
```

emerkle826 (changes requested):
3 minor fixes (2 for unused imports)

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/FeatureSet.java
```diff
@@ -23,6 +23,8 @@ public enum Feature {
     ASYNC_SCRUB_TASK("async_scrub_task"),
     ASYNC_MOVE_TASK("async_move_task"),
     ASYNC_GC_TASK("async_gc_task"),
+    RELOAD_INTERNODE_TRUSTSTORE("reload_internode_truststore"),
+    GET_INTERNODE_TRUSTSTORE("get_internode_truststore"),
```
burmanm, line 27: This should probably be removed, only the reload_internode_truststore is valid.

emerkle826: Good catch

management-api-agent-dse-6.8/src/main/java/com/datastax/mgmtapi/shim/DseAPI68.java
Miles-Garnsey: Sure, done.

management-api-server/src/main/java/com/datastax/mgmtapi/resources/models/FeatureSet.java
Miles-Garnsey: Done.

emerkle826 (approved):

PR #528, opened 2024-08-29
Add Cassandra 4.1.6 to the build matrix

### LINKED ISSUE #527: Add Cassandra 4.1.6 to the build matrix
This will likely require a bit more than simply updating version numbers in the Docker release job. See https://github.com/k8ssandra/management-api-for-apache-cassandra/discussions/526

emerkle826:
Fixes #527


burmanm:
question: I wonder if we'll start needing separate 4.1 smoke tests soon? So that 4.1.2, 4.1.3, 4.1.6 work as they have different paths.

emerkle826:
> question: I wonder if we'll start needing separate 4.1 smoke tests soon? So that 4.1.2, 4.1.3, 4.1.6 work as they have different paths.

I was thinking about that. What I want to do is have LifecycleIT run against the different versions as it should flush out these kinds of issues. I would really appreciate upstream not making some of these types of internal changes within a major-minor release, but they are internal bits that are normally not exposed, so it is what it is.

emerkle826:
> question: I wonder if we'll start needing separate 4.1 smoke tests soon? So that 4.1.2, 4.1.3, 4.1.6 work as they have different paths.

@burmanm I added https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/528/commits/2c3530290b356ee98cd37885833dd890d465ecc5 to test 4.1.2 and 4.1.5 to make sure we don't break things there.

burmanm (approved):

PR #536, opened 2024-09-10
Refactor CassandraDaemonInterceptor into SPI

emerkle826:
@burmanm This refactors the CassandraDaemonInterceptor so that we can implement it differently for 5.0_ agents, but it doesn't quite get us there. The problem with this is that the `CassandraDaemonInterceptor5x` implementation can't call the `Server.ConnectionTracker` constructor since it's private. To get around that, I'd have to refactor the `CassandraAPI` class to build the Socket initializer with something that is not a true Server.ConnectionTracker object, but rather a cloned implementation. I'm not sure I like that idea as that creates another bit of Cassandra internals that we implement and maintain separately.

### DIFF
--- a/management-api-agent-common/src/main/java/com/datastax/mgmtapi/interceptors/CassandraDaemonInterceptor.java
+++ b/management-api-agent-common/src/main/java/com/datastax/mgmtapi/interceptors/CassandraDaemonInterceptor.java
@@ -15,8 +15,11 @@ import io.netty.channel.EventLoopGroup;
 import io.netty.channel.epoll.Epoll;
 import io.netty.channel.epoll.EpollEventLoopGroup;
 import java.io.File;
+import java.lang.reflect.Constructor;
 import java.nio.file.Paths;
 import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


management-api-agent-5.0.x/src/main/java/com/datastax/mgmtapi/interceptors/CassandraDaemonInterceptor5x.java
```diff
@@ -0,0 +1,103 @@
...
+      logger.info("Starting DataStax Management API Agent for Apache Cassandra v0.1");
+
+      NodeOpsProvider.instance.get().register();
+
+      if (!Epoll.isAvailable()) throw new RuntimeException("Event loop needed");
+
+      File unixSock = Paths.get(socketFileStr).toFile();
+
+      final EventLoopGroup group = new EpollEventLoopGroup(8);
+      final Server.ConnectionTracker connectionTracker = new _ConnectionTracker();
```
emerkle826, line 67: This is the problematic bit. We can't call the new constructor that takes a BooleanSupplier function because it was marked private.

burmanm (approved):

PR #540, opened 2024-09-20
Add sstable-tools to Cassandra 4.1 images

burmanm:
Also, update the base builder image..


emerkle826 (approved):
Do we only want this on 4.1 images? Want to create a ticket to port it to other images and we can add it to them later? 
UPDATE: I see this is just a Draft for now....

burmanm:
Yeah, I created it only as a draft, I think we need to have a discussion around releasing the tool also. Otherwise it could block our releases (and always pulling from git and compiling feels weird).

PR #542, opened 2024-09-24
Move Spark directory creation from entrypoint to DSE Dockerfiles

### LINKED ISSUE #541: readOnlyRootFilesystem still fails due to some Spark folders in OpenShift
We've seen that OpenShift is still having some issues with the readOnlyRootFilesystem feature.
In the Dockerfile, we create additional directories for Spark: https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/dse/Dockerfile-dse6.8.jdk11#L171-L175

Then in the entrypoint.sh, we create subdirectories for Spark in some of the above directories: https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/dse/files/entrypoint.sh#L15-L18

While this works fine with "standard" Kubernetes distributions (including GKE), OpenShift isn't happy about it.
Standard distros seem to allow writes to these directories which OpenShift disallows.

We should remove the creation of these directories in the Dockerfile and rely on mounts at runtime instead, and have the entrypoint check for the presence of the directories before trying to create the additional subdirectories.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MAPI-66) by [Unito](https://www.unito.io)

emerkle826:
Fixes #541

Also addresses more potential readOnly root filesystem issues in Cassandra images

### DIFF
--- a/cassandra/Dockerfile-5.0.ubi8
+++ b/cassandra/Dockerfile-5.0.ubi8
@@ -34,21 +34,6 @@ RUN mkdir -m 775 ${CDC_AGENT_PATH} && \
   ln -s ${CDC_AGENT_EDITION}-${CDC_AGENT_VERSION}-all.jar cdc-agent.jar && \
   chmod -R g+w ${CDC_AGENT_PATH}
 
-FROM registry.access.redhat.com/ubi${UBI_MAJOR}/ubi-minimal:${UBI_BASETAG} AS cassandra-builder
-ARG URL_PREFIX=https://dlcdn.apache.org/cassandra
-ARG CASSANDRA_VERSION
-
-ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
-ENV CASSANDRA_HOME=/opt/cassandra
-
-# Update base layer
-RUN microdnf install --nodocs shadow-utils \
-    && groupadd -r cassandra --gid=999 \
-    && useradd -m -d "$CASSANDRA_HOME" -r -g cassandra -G root --uid=999 cassandra \
-    && microdnf update && rm -rf /var/cache/yum \
-# Install packages needed during install process
-    && microdnf install tar gzip unzip && microdnf clean all
-
 ###
 # Download Cassandra archive
 ###
@@ -141,8 +126,8 @@ RUN microdnf install --nodocs shadow-utils \
     && microdnf clean all
 
 # Copy trimmed installation
-COPY --from=cassandra-builder --chown=cassandra:root ${CASSANDRA_PATH} ${CASSANDRA_PATH}
-COPY --from=cassandra-builder --chown=cassandra:root ${CASSANDRA_PATH}/LICENSE.txt /licenses/
+COPY --from=builder --chown=cassandra:root ${CASSANDRA_PATH} ${CASSANDRA_PATH}
+COPY --from=builder --chown=cassandra:root ${CASSANDRA_PATH}/LICENSE.txt /licenses/
 COPY --from=builder --chown=cassandra:root ${CDC_AGENT_PATH} ${CDC_AGENT_PATH}
 COPY --from=mgmtapi-setup --chown=cassandra:root ${MAAC_PATH} ${MAAC_PATH}
 
--- a/cassandra/Dockerfile-5.0.ubi8
+++ b/cassandra/Dockerfile-5.0.ubi8
@@ -34,7 +34,7 @@ RUN mkdir -m 775 ${CDC_AGENT_PATH} && \
   ln -s ${CDC_AGENT_EDITION}-${CDC_AGENT_VERSION}-all.jar cdc-agent.jar && \
   chmod -R g+w ${CDC_AGENT_PATH}
 
-FROM --platform=linux/amd64 registry.access.redhat.com/ubi${UBI_MAJOR}/ubi-minimal:${UBI_BASETAG} AS cassandra-builder-amd64
+FROM registry.access.redhat.com/ubi${UBI_MAJOR}/ubi-minimal:${UBI_BASETAG} AS cassandra-builder
 ARG URL_PREFIX=https://dlcdn.apache.org/cassandra
 ARG CASSANDRA_VERSION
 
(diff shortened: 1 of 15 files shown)


emerkle826:
@burmanm This should handle _most_ of the read only stuff for DSE images. However, there is still a spot in the entrypoint that potentially does some stuff with JXM remote password here:
https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/dse/files/entrypoint.sh#L115-L123

There's two parts to that logic. The first part is checking to see if a `/config/jxmremote.password` file exists AND `$CASSANSDRA_CONF/jmxremote.password` does not exist. If that is the case, it creates a symlink from the one in `/conf` to the `$CASSANDRA_CONF` directory.

We can address that in the Dockerfile I suppose.

The second part of the logic, which is hooked by the first, then checks to see if the jmxremote.password file is specified in `cassandra-env.sh` via JVM_OPTS. In my testing, it seems that this is always the case for the DSE images. That is, this line always exists in cassandra-env.sh already, so the file never has this edit made since it is not needed. We can, however, move this logic to the Dockerfile as well.

Thoughts?

dse/Dockerfile-dse6.8.jdk11
```diff
@@ -191,6 +196,27 @@ RUN if ! grep -qxF "JVM_OPTS=\"\$JVM_OPTS -javaagent:${MGMT_AGENT_JAR}\"" < ${CA
...
+RUN if ! grep -qxf "JVM_OPTS=\"\$JVM_OPTS -Dinsights.default_mode=disabled\"" < ${CASSANDRA_CONF}/cassandra-env.sh ; then \
+    echo "" >> ${CASSANDRA_CONF}/cassandra-env.sh && \
+    echo "if [ ! -z \"\$MGMT_API_DISABLE_MCAC\" ]; then" >> ${CASSANDRA_CONF}/cassandra-env.sh && \
+    echo "  JVM_OPTS=\"\$JVM_OPTS -Dinsights.default_mode=disabled\"" >> ${CASSANDRA_CONF}/cassandra-env.sh && \
+    echo "fi" >> ${CASSANDRA_CONF}/cassandra-env.sh ; \
+  fi
+# link /config/jmxremote.password if it exists and there isn't one in CASSANDRA_CONF
+RUN if [ -e "/config/jmxremote.password" ] && [ ! -e "$CASSANDRA_CONF/jmxremote.password" ]; then \
+    ln -s /config/jmxremote.password $CASSANDRA_CONF/jmxremote.password ; \
+  fi
```
emerkle826, line 209: @burmanm Will this ever work?
The `/config` directory would be a volume mount in the case of cass-operator, right? So when the image is built, the Dockerfile will never see a `/config/jmxremote.password` file, right. I don't think moving this bit in here will work the same way it does in the entrypoint script, which isn't executed until after volume mounts are established.

burmanm: I'm not sure if we support remote JMX in DSE at all? So it wouldn't make any difference even if we removed this file entirely.

dse/Dockerfile-dse6.8.jdk11
```diff
@@ -169,10 +170,14 @@ COPY --chown=dse:root --from=dse-server-prep $DSE_AGENT_HOME $DSE_AGENT_HOME
 # Create folders
 RUN (for dir in /var/lib/cassandra \
                 /var/lib/spark \
+                /var/lib/spark/worker \
```
burmanm, line 173: I don't think these are required here. If anyone wants to use Spark, they have to create a mount /var/lib/spark that cass-operator then adds to the pod.

My idea was more like making entrypoint do the following:

```
if [ -d /var/lib/spark ]; then
  mkdir -p /var/lib/spark/worker
  mkdir -p /var/lib/spark/rdd
  mkdir -p /var/log/spark/worker
  mkdir -p /var/log/spark/master
fi
```

And not manually create /var/lib/spark at all. That way, user needs to add PVC (or EmptyVolume I guess, but that would feel weird) for the Spark data if they wish to use Spark.

burmanm: Because why would anyone store Spark files to the pod's image filesystem?

emerkle826: Would we need to even create the spark sub directories? If `/var/lib/spark` is going to be a volume mount, there's no point in creating the sub directories in the Dockerfile as they will be masked by the volume mount, correct? So really, I think we can revert this part as only `/var/lib/spark` and `/var/log/spark` need to exist so that a volume can be mounted there later.

dse/Dockerfile-dse6.8.jdk11
emerkle826: Ok, I'll do that.

dse/Dockerfile-dse6.8.jdk11
burmanm: No, we couldn't do it in the Dockerfile, but we can do it in the entrypoint (as previously). I think we should just not create /var/lib/spark at all perhaps? If it appears as mounted dir, then we create those subdirs.

Assuming there's even any need to create them. Why wouldn't Spark itself create them?

emerkle826: Ok, I see. So if the parent directory exists, we assume it's a mounted volume and we can create the subdirs then. And remove the parent directory creation from the Dockerfile. So the only way the parent directories exist is if they are mounted, which wouldn't cause issue with a read-only root filesystem.

cassandra/Dockerfile-5.0.ubi8
```diff
@@ -34,6 +34,55 @@ RUN mkdir -m 775 ${CDC_AGENT_PATH} && \
...
+# Download Cassandra archive
+###
+RUN if test ! -e apache-cassandra-${CASSANDRA_VERSION}-bin.tar.gz; then curl -L -O "https://archive.apache.org/dist/cassandra/${CASSANDRA_VERSION}/apache-cassandra-${CASSANDRA_VERSION}-bin.tar.gz"; fi && \
+    tar --directory ${CASSANDRA_HOME} --strip-components 1 --extract --gzip --file apache-cassandra-${CASSANDRA_VERSION}-bin.tar.gz && \
+    chown -R cassandra:root ${CASSANDRA_HOME} && \
+    chmod -R a+rwX ${CASSANDRA_HOME}
+# MCAC isn't supported on ARM achitectures
+ENV CASSANDRA_PATH=${CASSANDRA_HOME}
+ENV CASSANDRA_CONF=${CASSANDRA_PATH}/conf
+ENV MCAC_PATH=/opt/metrics-collector
```
burmanm, line 62: We don't want to ship MCAC with 5.0 builds, it's not going to work.

burmanm: I'm not sure if it works with most 4.1 versions either, but for sure not in 5.0

emerkle826: I reverted all the changes to the 5.0 Dockerfile since we don't need them if we will never add MCAC to the image. The removal has already happened in the entrypoint script. I'll leave it in for the 4.1 images, though I think you a re right in that it doesn't work in 4.1.

burmanm (approved):

PR #554, opened 2024-10-23
Change CassandraTaskExports filtering

### LINKED ISSUE #553: Major compaction tracking metric is broken in 5.0.2
Currently, we do cast to a OperationType, however this approach will not work in the extended metrics anymore as we use an older version of the class in agent-common. This causes major compactions to be filtered out from metrics due to a IllegalArgumentException (this Exception is not printed in the logs).

In 5.0.2, we need a newer version or we need to simply do this filtering through some string handling to get what we want.

burmanm:
Uses String instead of OperationType for compare

And fix CHANGELOG ordering.

Fixes #553

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,9 @@ Changelog for Management API, new PRs should update the `main / unreleased` sect
 ```
 
 ## unreleased
-* [ENHANCEMENT] [#552](https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/552) Improve "liveness" probe implementation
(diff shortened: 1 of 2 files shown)


emerkle826 (changes requested):
Get rid of the unused import and I think things look good.

emerkle826 (approved):
Looks fine. Feel free to merge when the tests run as long as there aren't consistent failures around the OperationType stuff.

burmanm:
The failed tests are just timeouts when trying to even start the test (and some of them keep hanging).

PR #561, opened 2024-10-28
Fix LatencyMetrics parsing to take into account that getCount() retur…

### LINKED ISSUE #560: LatencyMetrics processing is incorrect in certain Cassandra versions
We set our buckets to accommodate for nanoseconds in latency metrics, however this isn't true anymore in newer versions of Cassandra, which are actually storing the dropwizard with microseconds. 

Thus, we need to modify the timerFiller with nanosecond buckets for 3.11/4.0, but microseconds for 4.1/5.0.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MAPI-77) by [Unito](https://www.unito.io)
┆Issue Number: MAPI-77

burmanm:
…ns different reservoir values for 3.11/4.0 and 4.1/5.0. In latter, the reservoir stores in microseconds while in the first case, they are nanoseconds.

Fixes #560


emerkle826 (approved):
This looks fine. As long as some of the MetricsIT tests pass, we can merge this

PR #591, opened 2025-01-17
Fix Artifact release to also publish HCD 1.2 Agent

### LINKED ISSUE #586: Optimize Docker image releases
Currently, when a new version of Management API is released, the Docker image build/push process rebuilds the Management API jars for each combination of supported server versions in the README.md, which is growing unbounded at the moment. There are 2 issues with this approach:

1. We are building images that aren't actually using the released artifacts (though they are built from the exact same code)
2. The image builds are slower than they need to be and often times have to be restarted do to errors like this:

```
 #24 ERROR: process "/bin/sh -c cd /tmp && mvn -q -ff -T 1C install -Dskip.surefire.tests -DskipTests -DskipOpenApi -P dse && rm -rf target" did not complete successfully: exit code: 1
------
 > [mgmtapi-setup 13/24] RUN cd /tmp && mvn -q -ff -T 1C install -Dskip.surefire.tests -DskipTests -DskipOpenApi -P dse && rm -rf target:
Error: ERROR] Failed to execute goal on project datastax-mgmtapi-agent-dse-6.8: Could not resolve dependencies for project io.k8ssandra:datastax-mgmtapi-agent-dse-6.8:jar:0.1.0-SNAPSHOT: Failed to collect dependencies at com.datastax.dse:dse-core:jar:6.8.53 -> com.datastax.dse:dse-db:jar:6.8.53 -> io.netty:netty-all:jar:4.1.100.1.dse -> io.netty:netty-handler-ssl-ocsp:jar:4.1.100.1.dse: Failed to read artifact descriptor for io.netty:netty-handler-ssl-ocsp:jar:4.1.100.1.dse: Could not transfer artifact io.netty:netty-handler-ssl-ocsp:pom:4.1.100.1.dse from/to artifactory-releases (https://repo.datastax.com/datastax-releases-local): /root/.m2/repository/io/netty/netty-handler-ssl-ocsp/4.1.100.1.dse/netty-handler-ssl-ocsp-4.1.100.1.dse.pom.part (No such file or directory) -> [Help 1]
```

We should adjust the image creation process so that it depends on the release artifacts and simply uses them as opposed to rebuilding them for each server/platform combination.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MAPI-94) by [Unito](https://www.unito.io)
┆Issue Number: MAPI-94

### LINKED ISSUE #590: Fix Artifact release to also publish HCD 1.2 Agent
With the v0.1.91 Release, a new HCD 1.2 agent was added, but the artifact was not included in the Release zip bundle.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MAPI-96) by [Unito](https://www.unito.io)
┆Issue Number: MAPI-96

emerkle826:
* Combine release jobs into single workflow file
* Refactor Integration tests into its own submodule. Refactor Dockerfiles so building an image does not have to rebuild the artifacts within the Docker image build process
* Update Docker release to use release version artifacts
Fixes #590
Fixes #586


emerkle826:
@burmanm I would appreciate a review of this if you have some time.

dse/Dockerfile-dse6.8.jdk8
```diff
@@ -19,16 +19,14 @@ ARG DOWNLOAD_URL=${URL_PREFIX}/${TARBALL}
 ARG DSE_AGENT_VERSION=6.8.39
 ARG DSE_AGENT_TARBALL=datastax-agent-${DSE_AGENT_VERSION}.tar.gz
 ARG DSE_AGENT_DOWNLOAD_URL=${URL_PREFIX}/${DSE_AGENT_TARBALL}
-
+ADD ${DOWNLOAD_URL} /${TARBALL}
```
burmanm, line 22: Whitespace might not work correctly?

burmanm:
I assume the main benefit is faster CI when we don't need to compile mgmt-api for every version all the time?

emerkle826:
> I assume the main benefit is faster CI when we don't need to compile mgmt-api for every version all the time?

Yes, but really it's an attempt to get more reliable Docker image builds. Lately, when we do a release, I've had to re-run some of the Docker builds multiple times because each of the 118 images will also run a Maven build. Eventually a handful of them will fail because of some issue pulling artifacts from our Maven Artifactory.

With these changes, we build the branch/tag once and cache the build artifacts so the follow-on jobs can just restore and then build images by just copying the artifacts in.

Another approach I considered was going back to a "base" image that just builds the artifacts, and then pushing that image to a local repo that all the real image builds could simply pull from. I could still do that, if you think that would be a better approach.

dse/Dockerfile-dse6.8.jdk8
emerkle826: I guess when I put the `ADD` line, I just put it in place of an existing blank line. Using `ADD` here seems to allow for better layer cache hits as opposed to using a `RUN` command that downloads the bundle.

Also, for the Management API artifact layers, if we aren't building the code in the image, we can get away with a tiny Alpine image. The extra benefit of `ADD` here is that we don't need a base image that has `curl` or `wget` just to pull the bundles and unzip them. So I use the `ADD` lines there for that benefit as well.

burmanm: I didn't mean the whitespace line, but the whitespace in the ADD command. This will result in incorrect URL with encoded whitespace between ${DOWNLOAD_URL} and the `/`character.

burmanm:
> I could still do that, if you think that would be a better approach.

No, this is fine, we get the built files in the action artifacts, easier to debug if something went wrong.

burmanm (approved):

PR #594, opened 2025-01-28
Update HCD 1.2 Agent for CC 5.0.2

### LINKED ISSUE #593: Update HCD 1.2 Agent for CC 5.0.2
Converged Core `5.0.2-94c48ebf299a` was released and pulls in some Cassandra `Dispatcher` class changes that require the HCD 1.2 agent to be updated, similar to changes that went in for the Cassandra 4.1 agent.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MAPI-97) by [Unito](https://www.unito.io)
┆Issue Number: MAPI-97

emerkle826:
Fixes #593

### DIFF
--- a/management-api-agent-hcd-1.2.x/src/main/java/org/apache/cassandra/transport/UnixSocketServerHcd.java
+++ b/management-api-agent-hcd-1.2.x/src/main/java/org/apache/cassandra/transport/UnixSocketServerHcd.java
@@ -97,7 +97,34 @@ public class UnixSocketServerHcd {
         // logger.info("Executing {} {} {}", request, connection.getVersion(),
         // request.getStreamId());
 
-        Message.Response r = request.execute(qstate, queryStartNanoTime);
+        // In Cassandra 4.1.6, CASSANDRA-19534 changed the Message.Request.execute method signature
+        // to take a new Dispatcher.RequestTime object instead of a primitive long for the Query
+        // start time. We'll need to introduce reflection here to create the correct Objects and
+        // make the correct calls based on which version of 4.1.x we are. For Cassandra 5.0, this
+        // patch was ported between 5.0-rc1 and 5.0-rc2.
+        Message.Response r = null;
+        try {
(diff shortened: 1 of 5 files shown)


management-api-agent-hcd-1.2.x/src/main/java/org/apache/cassandra/transport/UnixSocketServerHcd.java
```diff
@@ -97,7 +97,34 @@ protected void channelRead0(ChannelHandlerContext ctx, Message.Request request)
...
+              dispatcherRequestTime.getDeclaredMethod("forImmediateExecution");
+          Method requestExecute =
+              Message.Request.class.getDeclaredMethod(
+                  "execute", QueryState.class, dispatcherRequestTime);
+          r =
+              (Message.Response)
+                  requestExecute.invoke(request, qstate, forImmediateExecution.invoke(null));
+
+        } catch (ClassNotFoundException cfne) {
+          logger.debug(
```
burmanm, line 121: I guess we will never need this part of the code in HCD? Or in 5.0.

emerkle826: Probably not, unless an HCD 1.2 image is built with a Converged Core 5 version that has specific commits that make the Dispatcher method signatures match this stuff here.
Honestly, this reflection stuff is not great, but the only alternative is to duplicate the code out into more agent submodules for specific point releases of the C*/DSE/HCD server code. They are pretty good at not changing the API within a <major>.<minor> release, but this isn't exactly public APIs we are using. But I digress....

emerkle826: To add to this, the main reason I'm leaving this code in place is in the case HCD needs to release a new 1.2 version and my patch for it to upgrade the Converged Core version does not get merged before the release. Then we would be in a situation where the Agent is using a Dispatcher signature that doesn't exist in the HCD build.

PR #609, opened 2025-02-19
Refactor HCD Agents

### LINKED ISSUE #608: Refactor HCD Agents
The Agents for HCD have been under some fluctuation lately due to Converged Cassandra changes. Currently, HCD 1.1 is based on CC 4, while HCD 1.2 is based on CC 5. However, HCD 1.2 will soon switch back to CC 4. All of the underlying CC versions and changes create compilation and runtime issues for the Agent, which is why we have different Agents for Cassandra 4.0, 4.1 and 5.0, as well as DSE 6.8 and DSE 6.9. Since HCD 1.1 and 1.2 are now going to be CC 4 based, it might be better to refactor the HCD agent modules in this project to reflect the CC version as opposed to the HCD version.

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MAPI-104) by [Unito](https://www.unito.io)
┆Issue Number: MAPI-104

emerkle826:
Fixes #608


emerkle826:
@burmanm If you get a chance, I'd appreciate a review here. Mostly just a refactoring of the HCD Agents, and a CC 4 version upgrade that brings about internal API changes....

burmanm:
I'm trying to understand certain things. Is cc4 incompatible with our -4.0.x agent and is cc5 going to be incompatible with our -5.0.x ?

Just thinking if we really need even more submodules, since we might miss a fix that would apply to multiple modules at this point.

management-api-agent-hcd-cc4/src/main/java/org/apache/cassandra/transport/UnixSocketServerHcd.java
```diff
@@ -284,18 +319,64 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> ou
...
+                        StartupMessage.class,
+                        ClientResourceLimits.Overload.class);
+                Message.Response response =
+                    (Message.Response)
+                        processRequest.invoke(
+                            null,
+                            (ServerConnection) connection,
+                            startup,
+                            ClientResourceLimits.Overload.NONE);
+                if (response.type.equals(Message.Type.AUTHENTICATE))
```
burmanm, line 364: For safety, this should be with { } instead of relying on indentation and a single line response. We don't use this type of "old bastard C coding who like bugs" in management-api elsewhere.

Btw, in this PR there are 3 ways of writing the same code, which feels weird. First the normal, then one with shorter if else conditional and then this one without {}.

emerkle826: Good catch. I'll clean it up

emerkle826: Actually, I'm glad you brought this up because I missed the same "if message type is AUTHENTICATE" in the code block above.

burmanm (approved):

PR #616, opened 2025-03-05
Create notificationListener before any repairs are submitted

burmanm:
Modify the agent to use a single listener for the repair progress status updates and create that listener before any repair jobs are scheduled

### DIFF
--- a/management-api-agent-common/src/main/java/com/datastax/mgmtapi/util/JobExecutor.java
+++ b/management-api-agent-common/src/main/java/com/datastax/mgmtapi/util/JobExecutor.java
@@ -5,18 +5,100 @@
  */
 package com.datastax.mgmtapi.util;
 
+import com.datastax.mgmtapi.ShimLoader;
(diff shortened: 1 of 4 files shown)


management-api-agent-common/src/main/java/com/datastax/mgmtapi/util/JobExecutor.java
```diff
@@ -5,19 +5,101 @@
...
+                      String[] split = ((String) notification.getSource()).split(":");
+                      if (split.length > 1) {
+                        Integer.parseInt(split[1]);
+                        return true;
+                      }
+                    } catch (NumberFormatException e) {
+                      return false;
+                    }
+                  }
+                  return false;
```
adejanovski, line 98: **praise:** While I haven't been able to reproduce the issues we've seen with lost repair tracking, I think this is a solid improvement to make the code resilient to some errors that could break repair tracking.

adejanovski (approved):

PR #621, opened 2025-03-24
Update Dependencies

### LINKED ISSUE #620: Update Dependencies
Netty, Jackson and Guava dependencies should be updated to address security concerns and CVEs

 Netty -> 4.1.118.Final
Jackson -> 2.18.3
Guava -> 33.4.0-jre

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MAPI-110) by [Unito](https://www.unito.io)
┆Issue Number: MAPI-110

emerkle826:
- Update Netty to 4.1.118.Final
- Update Jackson to 2.18.3
- Update GUava to 33.4.0-jre
- Update Maven plugin versions to accomodate

Fixes #620

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changelog for Management API, new PRs should update the `main / unreleased` sect
 ```
 
 ## unreleased
+* [ENHANCEMENT] [#620](https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/620) Update Dependnecies
 * [BUGFIX] [#438](https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/438) Add Cassandra 5.1 Agent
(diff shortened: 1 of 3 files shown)


emerkle826:
@burmanm Any issue with updating these? The Netty and Guava libraries are only explicitly used in the server part of management API (separate JVM from C*/DSE). The Jackson library is shaded in the Agents so that there aren't any version conflicts with what C*/DSE provide. The tests look good in here ( a couple failing that are still flaky)

burmanm:
Should be fine if the APIs for dataformat hasn't changed. As long as the unit tests check the dataformat usage then I have no issues. Assuming we don't inherit some other version for tests..

emerkle826:
It looks like all the tests are passing. In the past, Jackson compatibility issues usually showed up, at the latest, during startup tests and all of those are passing.

burmanm (approved):

PR #627, opened 2025-03-31
Bump MCAC to 0.3.6

rzvoncek:


cassandra-trunk/Dockerfile-trunk.ubi8
```diff
@@ -40,6 +40,7 @@ ARG COMMITSHA="HEAD"
 ARG CASSANDRA_BRANCH="trunk"
 ENV CASSANDRA_PATH=/opt/cassandra
 ENV CASSANDRA_FILES_PATH=/opt/cassandra_files
+RUN apt-get update && apt-get install -y python3
```
emerkle826, line 43: :suggestion

Can we simply move this to line 49 below and include it with the list of things being installed there? So that line would look like:
```
&& apt-get install -y --no-install-recommends git ant ant-optional make python3 \
```

emerkle826 (approved):
This looks good, but it would be better if we can consolidate the Python3 install step with the RUN line that already installs things for the Cassandra trunk Dockerfile

cassandra-trunk/Dockerfile-trunk.ubi8
rzvoncek: oh I totally missed this, sure, good one.

emerkle826 (approved):
:+1:

PR #637, opened 2025-04-24
Add Cassandra 5.0.4 to the build matrix

### LINKED ISSUE #634: Fix typo in Readme file
Fix typo in Readme file [Gogle to Google]

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MAPI-116) by [Unito](https://www.unito.io)
┆Issue Number: MAPI-116

### LINKED ISSUE #635: Add Cassandra 5.0.4 to the build matrix
┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MAPI-117) by [Unito](https://www.unito.io)
┆Issue Number: MAPI-117

### LINKED ISSUE #636: Add JDK17 for Cassandra 5.0 images
┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MAPI-118) by [Unito](https://www.unito.io)
┆Issue Number: MAPI-118

emerkle826:
Also updates Cassandra 5.0 images to use JDK17

Fixes #635
Fixes #636
Fixes #634

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -509,9 +509,9 @@ jobs:
       fail-fast: false
       max-parallel: 4
       matrix:
-        cassandra-version: [5.0.1, 5.0.2, 5.0.3]
+        cassandra-version: [5.0.1, 5.0.2, 5.0.3, 5.0.4]
         include:
-          - cassandra-version: 5.0.3
+          - cassandra-version: 5.0.4
             latest: true
     runs-on: ubuntu-latest
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Changelog for Management API, new PRs should update the `main / unreleased` sect
 ```
 
 ## unreleased
+* [FEATURE] [#635](https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/635) Add Cassandra 5.0.4 to the build matrix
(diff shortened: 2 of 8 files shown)


cassandra/scripts/docker-entrypoint.sh
bradfordcp, line 1: Should this entry point script work for both JDK11 and JDK17? I see the new file named `...-jdk8.sh` and I'm left wondering if it would be better to have one per java version.

emerkle826: > Should this entry point script work for both JDK11 and JDK17? I see the new file named ...-jdk8.sh and I'm left wondering if it would be better to have one per java version.

The one with "jdk8" in the name is only for Cassandra 3.11, becazuse it's the only one that installs more than one JDK.

The regular entrypoint script will work for whatever JDK is installed as default, i.e. whatever would be used when just "java" is the command. So far, the only images that have more than one JDK installed are the 3.11 images, and that is because the Management API server jar has to run in JDK11+, but Cassandra 3.11 only supports Java 8.

So I made the entrypoint script for 3.11 (which we don't even build images for anymore) "special", while the regular script just uses the default "java". If we want more than one JDK for Cassandra 5.0 images, then I'll have to do something similar for the entrypoint script for Cassandra 5.0

burmanm (approved):

PR #691, opened 2025-10-17
Add bootstrap state check for /probes/readiness

### LINKED ISSUE #690: Add check to /readiness that the node has finished bootstrapping
It is possible that our current /probes/readiness returns "OK" even if the node is not really ready to reply to requests yet. This is because system.local can be queried without bootstrapping being in COMPLETED state:

https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/cql3/QueryProcessor.java#L358

So instead of returning OK if system.local succeeds, we should check that bootstrapped state has reached COMPLETED at least once:

https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/SystemKeyspace.java#L1279

burmanm:
Fixes #690

### DIFF
--- a/management-api-server/src/main/java/com/datastax/mgmtapi/resources/K8OperatorResources.java
+++ b/management-api-server/src/main/java/com/datastax/mgmtapi/resources/K8OperatorResources.java
@@ -95,11 +95,13 @@ public class K8OperatorResources extends BaseResources {
   public Response checkReadiness() {
     return handle(
         () -> {
-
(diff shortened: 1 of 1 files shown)


emerkle826 (approved):
This looks good. Let's wait to see the tests run and if they pass, feel free to merge.

burmanm:
It passed my sniff test, so it should be fine.

PR #740, opened 2026-04-14
Fix Metrics endpoint for HCD 2.0/CC5 Agent

### LINKED ISSUE #739: Fix Metrics endpoint for HCD 2.0/CC5 Agent
When accessing the internal Metrics endpoint provided by the HCD Agent (for HCD 2.0), we get the following:
```
aused by: java.lang.ClassNotFoundException: io.netty.handler.codec.http.HttpServerCodec
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:527)
	... 30 common frames omitted
```

It seems the HCD Agent builds do not explictly include `netty-codec-http` like the [Cassandra 5.0 Agent does](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-agent-5.0.x/pom.xml#L40-L44). This inclusion is needed due to Cassandra 5 (and consequently CC5) having an [explicit exclusion of extra Netty libraries](https://github.com/apache/cassandra/blob/cassandra-5.0.7/.build/parent-pom-template.xml#L746-L816). The internal Metrics endpoint requires the `netty-codec-http` artifact, so we should ensure it is added explicitly for Agents

emerkle826:
This patch adds an explicit dependnecy on netty-codec-http, which is required for the Metrics endpoints to work. This dependency has been explictly excluded in Cassandra 5, so for Agents that are based on Cassandra 5 (and newer), the netty-codec-http artifact needs to be explictly added back in. The MetricsIT test should catch this, but none of the IT tests in this project are run for HCD agents.

Fixes #739

### DIFF
--- a/management-api-agent-6.0.x/pom.xml
+++ b/management-api-agent-6.0.x/pom.xml
@@ -17,7 +17,15 @@
   <artifactId>datastax-mgmtapi-agent-6.0.x</artifactId>
   <properties>
     <cassandra6.version>6.0-alpha2-SNAPSHOT</cassandra6.version>
-    <netty.http.codec.version>4.1.118.Final</netty.http.codec.version>
+    <!--
+    The version here needs to match the Netty version from Cassandra upstream.
+    So when the version above changes, this may need to change to stay in sync.
+    Cassandra 5 explicitly excludes some Netty libraries from its netty-all
+    dependnecy, but the Metrics endpoint relies on netty-http-codec, so we must
+    explicitly bring it back in, but keep the version the same as the transitive
+    netty-all that we get from Cassandra
+    -->
+    <netty.http.codec.version>4.1.130.Final</netty.http.codec.version>
   </properties>
   <dependencies>
     <!-- Need to explicitly declare SLF4J as "provided" to avoid it being bundled into the resulting agent jarfile -->
@@ -105,7 +113,6 @@
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>build-helper-maven-plugin</artifactId>
-        <version>1.8</version>
         <executions>
           <execution>
             <phase>initialize</phase>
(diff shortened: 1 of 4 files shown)


management-api-agent-6.0.x/pom.xml
```diff
@@ -17,7 +17,15 @@
   <artifactId>datastax-mgmtapi-agent-6.0.x</artifactId>
   <properties>
     <cassandra6.version>6.0-alpha2-SNAPSHOT</cassandra6.version>
-    <netty.http.codec.version>4.1.118.Final</netty.http.codec.version>
+    <!--
+    The version here needs to match the Netty version from Cassandra upstream.
```
burmanm, line 21: Do we need to add the netty-codec-http here also for the 6.x?

burmanm (approved):
Other than potential 6.x missing codec one, this should do the work..

management-api-agent-6.0.x/pom.xml
emerkle826: I'm not sure I understand the question.

We do need it for 6.0 (used to be 5.1-SNAPSHOT), and we have had it since then I think. The only thing I am changing here for 6.0 is the version so that it matches the Netty version that comes with 6.0 currently. It was set to 4.1.118.Final and I am bumping it here to 4.1.130.Final (along with the xml comment that will hopefully help devs keep this up to date going forward)

emerkle826: Not sure if you were asking about adding the `netty-code-http` dependency to 6.0 or not, but it is already in the pom [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/management-api-agent-6.0.x/pom.xml#L50-L54)

emerkle826:
> Other than potential 6.x missing codec one, this should do the work..

I have tested this locally by modifying the Test framework and adding an HCD 2.0 Dockerfile. It was just a quick and dirty attempt that only works locally as I already have an HCD 2.0.2 image loaded in my local Docker setup. I then ran the MetricsIT test to verify and the test passed locally. I have a TODO to get this test running against HCD, either in this repo or in the HCD code base. It needs to happen somewhere so things like this get caught sooner.

PR #765, opened 2026-06-15
Fix Agent for Cassandra trunk

### LINKED ISSUE #746: Fix Agent for Cassandra trunk
Issue #742 and PR #743 mostly address the build for the Agent for Cassandra trunk, but the naming of the Agent being `6.0.x` doesn't line up inn  a clear way. We should refactor the Agent and Maven sub-module so that it is more clear.

Since Cassandra 6.0 isn't GA yet (most recent is `6.0-alpha2`), maybe the `6.0.x` Agent stays and we add yet another `trunk` agent. Details to be fleshed out in the implementation.

emerkle826:
This patch fixes the class names of the Cassandra 6.0 Agent module as well as fixing the ObjectSerializer failures for both 6.0 and trunk (7.0).

Fixes #746

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -153,6 +153,11 @@ jobs:
     name: Build Cassandra trunk image
     runs-on: ubuntu-latest
     needs: run-unit-tests
+    strategy:
+      fail-fast: false
+      max-parallel: 2
+      matrix:
+        cassandra-version: ['6.0', 'trunk']
     steps:
       - name: Restore build workspace
         uses: actions/cache/restore@v5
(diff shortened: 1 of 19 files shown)


emerkle826:
@burmanm I'd appreciate a review if you have time. I've fixed an issue with the Agent for trunk (and 6.0-alpha2) and now we have a little more separation between 6.0 and 7.0 (assuming 6.0 will GA in the near future). The Dockerfile for 6.0 is very close to the "trunk" Dockerfile, so you may want to just look at the diff for the "trunk" Dockerfile to get an idea of what I'm changing.

The most notable is:
- Upgrading the base image to UBI 10
- Installing Adoptium/Temurin JDK 17 in these images as UBI 10 no longer provides anything less than JDK 21

UPDATE:
- UBI 10 only has Python 3.12 +, and cqlsh still requires 3.6 - 3.11, so I've dropped these back down to UBI9
- I left in the Temurin JDK, however _not_ installing `java-17-openjdk-headless` (like we do in the 5.0.x UBI image) means we don't get Python 3.9 installed by default (alonng side Python 3.11), and for some reason don't get a `usr/bin/python` file or symlink. So I have added that as well now, which gets cqlsh working in these images

PR #767, opened 2026-06-22
Publish the DSE images to a private icr.io registry

adejanovski:
Refactor CI/CD: Migrate DSE images to private ICR registry and eliminate artifacts

Changes:

CI Workflow (.github/workflows/ci.yaml):
- Merged build-dse-ubuntu-docker-images, build-dse-ubi-docker-images, and
  test-dse jobs into a single build-and-test-dse job
- Build and test now run in the same job/runner, eliminating the need for
  artifact-based image transfer between jobs
- Images are built, tested, and discarded within each test job
- Updated publish-dse job to rebuild images and push directly to
  icr.io
- Removed all artifact upload/download steps for DSE images

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -207,78 +207,11 @@ jobs:
           path: /tmp/cassandra.${{ matrix.cassandra-version }}.ubi.tar
           retention-days: 3
           
-  build-dse-ubuntu-docker-images:
-    name: Build DSE Ubuntu images
+  build-and-test-dse:
+    name: Build and Test DSE ${{ matrix.itTest }} [${{ matrix.dse-version }}, ${{ matrix.base-platform }}]
     runs-on: ubuntu-latest
     needs: run-unit-tests
-    strategy:
-      fail-fast: false
-      max-parallel: 8
-      matrix:
-        dse-version: ['6.8', '6.9']
-        base-platform: ['jdk8', 'jdk11']
-        exclude:
-          - dse-version: '6.9'
-            base-platform: 'jdk8'
-    steps:
-      - name: Restore build workspace
-        uses: actions/cache/restore@v5
-        with:
-          path: /home/runner/work/management-api-for-apache-cassandra/management-api-for-apache-cassandra
-          key: ${{ runner.os }}-workspace-${{ github.sha }}-${{ hashFiles('**/pom.xml') }}
-      - name: Set up JDK 11
-        uses: actions/setup-java@v5
-        with:
-          java-version: 11
-          distribution: 'zulu'
-      - name: Cache Maven packages
-        uses: actions/cache@v5
-        with:
-          path: ~/.m2
-          key: ${{ runner.os }}-m2-${{ github.sha }}-${{ hashFiles('**/pom.xml') }}
-          restore-keys: ${{ runner.os }}-m2
-      # Setup metadata based on the commit/tag that will be used for tagging the image
-      # Only build and publish a commit based tag
-      - name: Setup Docker meta
-        id: meta
-        uses: docker/metadata-action@v6
-        with:
-          images: datastax/dse-mgmtapi-6_8
-          tags: type=sha,format=long,prefix=${{ matrix.dse-version }}-${{ matrix.base-platform }}-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v4
-        with:
-          image: tonistiigi/binfmt:qemu-v7.0.0-28
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v4
-      - name: Login to DockerHub
-        uses: docker/login-action@v4
-        with:
-          username: ${{ secrets.DOCKER_HUB_USERNAME }}
-          password: ${{ secrets.DOCKER_HUB_PASSWORD }}
-      - name: Build DSE ${{ matrix.dse-version }}-${{ matrix.base-platform }}
-        id: docker_build
-        uses: docker/build-push-action@v7
-        with:
-          file: dse/Dockerfile-dse${{ matrix.dse-version }}.${{ matrix.base-platform }}
-          context: .
-          push: false
-          tags: ${{ steps.meta.outputs.tags }}
-          platforms: linux/amd64
-          target: dse
-          outputs: type=docker,dest=/tmp/dse.${{ matrix.dse-version }}.${{ matrix.base-platform }}.tar
-      - name: Upload Docker Image for DSE ${{ matrix.dse-version }}-${{ matrix.base-platform }}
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -207,25 +207,30 @@ jobs:
...
-  build-dse-ubuntu-docker-images:
-    name: Build DSE Ubuntu images
+  build-and-test-dse:
+    name: Build and Test DSE ${{ matrix.itTest }} [${{ matrix.dse-version }}, ${{ matrix.base-platform }}]
     runs-on: ubuntu-latest
     needs: run-unit-tests
+    # need a local docker registry to push images into for UBI builds
+    services:
+      registry:
+        image: registry:2
```
burmanm: registry:3

.github/workflows/ci.yaml
```diff
@@ -207,25 +207,30 @@ jobs:
...
       matrix:
         dse-version: ['6.8', '6.9']
-        base-platform: ['jdk8', 'jdk11']
+        base-platform: ['jdk8', 'jdk11', 'ubi']
+        itTest : ['LifecycleIT', 'KeepAliveIT', 'NonDestructiveOpsIT', 'DestructiveOpsIT', 'DSESpecificIT', 'NonDestructiveOpsResourcesV2IT', 'DockerImageIT', 'AsyncRepairIT', 'PortOverrideIT', 'MetricsIT']
         exclude:
           - dse-version: '6.9'
             base-platform: 'jdk8'
+          - dse-version: '6.8'
+            base-platform: 'jdk11'
```
burmanm, line 231: question: Why is 6.8 defaulting to jdk11 and then 6.9 to jdk8?

.github/workflows/ci.yaml
```diff
@@ -812,10 +731,16 @@ jobs:
...
-    name: Publish DSE 6.x Ubuntu image
+    name: Publish DSE 6.x image to ICR
     if: ${{ github.ref == 'refs/heads/master' && github.event_name == 'push'}}
-    needs: test-dse
+    needs: build-and-test-dse
     runs-on: ubuntu-latest
+    # need a local docker registry to push images into for UBI builds
+    services:
+      registry:
+        image: registry:2
```
burmanm: :3

.github/workflows/release.yaml
```diff
@@ -273,23 +274,24 @@ jobs:
...
-            --tag datastax/dse-mgmtapi-6_8:6.8-${{ matrix.image-base }} \
-            --tag datastax/dse-mgmtapi-6_8:6.8-ubi8 \
-            --tag datastax/dse-mgmtapi-6_8:${{ matrix.dse-version }}-${{ matrix.image-base }} \
-            --tag datastax/dse-mgmtapi-6_8:${{ matrix.dse-version }}-ubi8 \
-            --tag datastax/dse-mgmtapi-6_8:${{ matrix.dse-version }}-${{ matrix.image-base }}-$RELEASE_VERSION \
-            --tag datastax/dse-mgmtapi-6_8:${{ matrix.dse-version }}-ubi8-$RELEASE_VERSION \
+            --tag icr.io/ds-mission-control-private/dse-mgmtapi-6_8:6.8-${{ matrix.image-base }} \
+            --tag icr.io/ds-mission-control-private/dse-mgmtapi-6_8:6.8-ubi8 \
+            --tag icr.io/ds-mission-control-private/dse-mgmtapi-6_8:${{ matrix.dse-version }}-${{ matrix.image-base }} \
+            --tag icr.io/ds-mission-control-private/dse-mgmtapi-6_8:${{ matrix.dse-version }}-ubi8 \
```
burmanm, line 302: suggestion: I don't think we need to ship -ubi8 anymore. That was kept to retain backwards compatibility with older setups, but as they break anyway there's no reason to keep it anymore. The images themselves are -ubi9, but I think we can just stick to -ubi only.

.github/workflows/release.yaml
```diff
@@ -273,23 +274,24 @@ jobs:
...
+      - id: docker-login-icr
         uses: docker/login-action@v4
         with:
-          username: ${{ secrets.DOCKER_HUB_USERNAME }}
-          password: ${{ secrets.DOCKER_HUB_PASSWORD }}
+          username: iamapikey
+          password: ${{ secrets.PRIVATE_ICR_API_KEY }}
+          registry: icr.io
       - if: ${{ matrix.latest }}
         name: Publish ${{ matrix.dse-version }} (ubi) to Registry
```
burmanm, line 294: issue: So, there's a small issue here. The UBI Dockerfiles have these as their base:

ARG DSE_BASE_IMAGE=datastax/dse-mgmtapi-6_8:${DSE_VERSION}-jdk11

From which they start their build on. Now, since we're moving publishing this will no longer be happening. What we need here is to use the previously created icr.io Ubuntu builds as the base, otherwise our base will never receive newer versions.

This applies to the other DSE 6.9 part also, but I'll write this just here. 

Likewise, I'm assuming this will stop functioning in the Ubuntu images:

FROM datastax/dse-server:${DSE_VERSION} AS dse-server-prep

So we will not get updated stuff from the Ubuntu repo either. We need to modify these bases to use proper sources from now on.

.github/workflows/ci.yaml
emerkle826: This is an exclusion, so it's the other way around. DSE 6.9 is excluding jdk8, DSE 6.8 is excluding jdk11

.github/workflows/release.yaml
adejanovski: We're not moving the dse-server ones yet, but that'll happen shortly.
Let me fix the Dockerfiles.

.github/workflows/release.yaml
adejanovski: the 6.8 ubi images are definitely still using ubi8, so let's defer for now. I don't want to go down a rabbit hole.

.github/workflows/release.yaml
adejanovski: Fixed

.github/workflows/ci.yaml
adejanovski: fixed

.github/workflows/ci.yaml
adejanovski: fixed

.github/workflows/release.yaml
burmanm: That is irrelevant, it's a tag. It was used because we shipped to users the ImageConfig with "-ubi8" as suffix and changing the tag would have broken existing workflows.

This will break existing workflows so the -ubi8 suffix does not matter anymore, users will have to change their settings anyway.

burmanm (approved):

PR #776, opened 2026-07-22
Fix metric registry issues

### LINKED ISSUE #774: Metrics bugs related to recreating tables
There's a bug in metrics implementation in the registration phase. If we create a new table, remove it and then readd it, we keep the old reference alive in our metrics instead of the new one. This is because we failed to correctly remove the old one from our cache and the new one is seen as duplicate of the previous one due to missing comparison.

There's also one small concurrency bug in the familyCache adding, although I've not seen this in the any other instance than tests bombing that function - Cassandra itself does not seem to initialize that codepath in concurrent fashion.

And a small bucket bug, where we can have one-off errors.

burmanm:
This PR can be reviewed per commit, each commit is a complete feature (other than tests being added in the commit 3 for the first 2 commits).

Fixes several issues in the metrics library related to registering and removing the metrics as well as one bucket bug which has actually been in the MCAC already (as that calculation was lifted from there).

Fixes #774

### DIFF
--- a/management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricDefinition.java
+++ b/management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricDefinition.java
@@ -94,26 +94,20 @@ public class CassandraMetricDefinition
 
   @Override
   public int hashCode() {
-    return Objects.hash(labelNames, labelValues, metricName);
+    return Objects.hash(labelNames, labelValues, metricName, dropWizardName);
   }
 
   @Override
   public String toString() {
-    return "CassandraMetricDefinition{"
-        + "labelNames="
-        + labelNames
-        + ", labelValues="
-        + labelValues
-        + ", metricName='"
-        + metricName
-        + '\''
-        + ", valueGetter="
-        + valueGetter
-        + ", keep="
-        + keep
-        + ", filler="
-        + filler
-        + '}';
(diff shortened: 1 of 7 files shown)


management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricDefinition.java
```diff
@@ -82,7 +88,8 @@ public boolean equals(Object o) {
     return keep == that.keep
         && Objects.equals(labelNames, that.labelNames)
         && Objects.equals(labelValues, that.labelValues)
-        && Objects.equals(metricName, that.metricName);
+        && Objects.equals(metricName, that.metricName)
+        && Objects.equals(dropWizardName, that.dropWizardName);
```
emerkle826, line 92: shouldn't you also add `dropwizardName` to the `hashcode()` method as well?

emerkle826 (changes requested):
I looked this over and it looks ok to me. Just need to update the `hashcode()` method I think.

management-api-agent-common/src/main/java/io/k8ssandra/metrics/builder/CassandraMetricDefinition.java
burmanm: Yeah, fixed.

emerkle826 (approved):

PR #777, opened 2026-07-22
430: Uniform support for epoll/kqueue

### LINKED ISSUE #430: Use Java NIO as fallback for Netty when epoll is not available
### Issue:
Management API was developed to run on Linux. It takes advantage of this for its Netty usage by essentially requiring epoll for the event queue. This makes fully testing Management API difficult for Mac users.
### Solution:
It should be possible to detect the absence of epoll and fall back to using Java's NIO. While this will result in a performance hit, it will allow users not on Linux/AMD64 platforms to test. Also, if Management API is run and falls back to Java NIO, a warning should be logged indicating the degraded performance  and that the condition should only be acceptable in labs/development/testing/etc.

asquare:
- abstracts sockets & event loops into class NativeTransport
- uses epoll if available, kqueue next and fallback to nio if possible
- logs warning if neither epoll nor kqueue is available
Fixes #430

### DIFF
--- a/management-api-agent-common/src/main/java/com/datastax/mgmtapi/interceptors/CassandraDaemonInterceptor.java
+++ b/management-api-agent-common/src/main/java/com/datastax/mgmtapi/interceptors/CassandraDaemonInterceptor.java
@@ -8,12 +8,11 @@ package com.datastax.mgmtapi.interceptors;
 import com.datastax.mgmtapi.NodeOpsProvider;
 import com.datastax.mgmtapi.ShimLoader;
 import com.datastax.mgmtapi.ipc.IPCController;
+import com.datastax.mgmtapi.ipc.NativeTransport;
 import com.google.common.collect.ImmutableMap;
 import io.k8ssandra.metrics.interceptors.MetricsInterceptor;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.EventLoopGroup;
-import io.netty.channel.epoll.Epoll;
-import io.netty.channel.epoll.EpollEventLoopGroup;
 import java.io.File;
 import java.lang.reflect.Constructor;
 import java.nio.file.Paths;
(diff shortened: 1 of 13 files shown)


emerkle826 (changes requested):
@asquare Thanks for taking a shot at this. The changes look ok, but we need to update the CI to actually test this stuff on ARM.

### CI Unit Tests
If you look [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/.github/workflows/ci.yaml#L13-L74) you'll see the basic build and unit test phase. It runs on `ubuntu-latest` which is an AMD64 platform. We need to update this to also run on a macos-arm runner (I think something like `runs-on: macos-15`) in parallel. However, the DSE and HCD stuff doesn't need to be run on MacOS ARM. So you can skip the `-P dse,hcd` stuff on that runner.

You should be able to refactor this into a matrix with the platform being one axis. Another axis could be the Maven profiles enabled (`-P dse,hcd` only from AMD, not for ARM)

### CI Integration test images
This part might get a little tricky. After build and Unit tests run, then we [build an image](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/.github/workflows/ci.yaml#L77-L150) for each of the OSS versions we run ITs against (4.0, 4.1 and 5.0). In theory, these images work for Linux ARM, but I don't know if they need changes to also build them for Mac ARM. The images are built once and archived for later to run ITs on.

We should again update this job so that it also builds the Mac ARM test images. It already is a matrix, so here I think we just need to add an axis with the platform

### CI Integration testing
Once we get images built and saved off, we now need to update the actual integration tests [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/blob/master/.github/workflows/ci.yaml#L362-L441). It does this:
- Pulls down the previously built image based on Cassandra version and tags it with a specific tag that the Tests will look for
- Runs a single IT in a matrix of tests
This should be just another matrix axis added for the platform, and keep the image tagging the same on ARM as the existing AMD images when you run them on a Mac ARM instance. If you get hung up on this part, let me know.

emerkle826:
@burmanm If you get some time, could you take a look here and see if you agree with my requested changes?

burmanm:
I disagree with the approach taken here in this main point that I see in few places:

```
Unix domain socket "
              + "communication with Cassandra will not be available. "
              + "This fallback is only acceptable in development, testing, or lab environments.");
```

If we can't create Unix Domain Socket, we should just simply quit. We can have an additional parameter that disables the unix domain socket, but this behavior should not be changed. Instead of debugging crashing containers I would suddenly have to debug multiple systems to find out that the system for some reason started in the wrong mode.

This is hard no from me to change this behavior. By default it has to start or we have to die.

asquare:
Updated - now exiting if native transport is not available.

The warnings came from this requirement in the issue -
> If Management API is run and falls back to Java NIO, a warning should be logged indicating the degraded performance and that the condition should only be acceptable in labs/development/testing/etc.

asquare:
Hey @burmanm - I made the change you asked for, anything else?

burmanm (approved):

emerkle826 (approved):
@asquare I'll go ahead and approve this. I would appreciate if you could refer to my original comments about testing and create another ticket and we can address the testing efforts later.

asquare:
Created issue https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/790 for CI updates

PR #778, opened 2026-07-23
Fix PR owner logic

emerkle826:
This patch fixes the logic in the CI job that tries to exclude DSE build and test for PRs that are from forks. Forked repos will not have access to the Repository secrets which are needed to build and test the DSE agents.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,10 +52,10 @@ jobs:
           </settings>
           EOF
       - name: Build with Maven and run unit tests (k8ssandra repo)
-        if: github.repository_owner == 'k8ssandra'
+        if: ${{ github.event.pull_request.head.repo.full_name == 'k8ssandra/management-api-for-apache-cassandra' }}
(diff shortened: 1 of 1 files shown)


burmanm (approved):
Difficult to see from tests if this helped as it came from the correct repo and not a fork, but I assume it should work..

emerkle826:
> Difficult to see from tests if this helped as it came from the correct repo and not a fork, but I assume it should work..

:facepalm: I should have tried to PR this from my own fork, that would have made a lot more sense....

emerkle826:
@burmanm there's still a problem with this in that it builds and tests DSE stuff when a PR is opened and the PR is from a non-forked branch (see https://github.com/k8ssandra/management-api-for-apache-cassandra/actions/runs/30017103890)

but this specific logic falls to "forked" when it's just a commit push to a branch that already has a PR, or a branch that is pushed and a PR hasn't been opened yet (see https://github.com/k8ssandra/management-api-for-apache-cassandra/actions/runs/30015778316)

I'll revisit this with some improvements, I don't want to merge it like it is just yet

